PR #133, opened 2015-02-19
bulk_create should return objects instead of None

coagulant:


Suor:
Merged, thanks. Regression test would be nice, though.

PR #134, opened 2015-02-25
First implementation of dog-pile effect avoidance via python-redis-lock additions

Karmak23:
Hi @Suor,

here is a patch to enable dog-pile avoidance in `cacheops`.

Could you please review it and see if I didn't miss anything ?

This patched version has been running for hours on my 1flow development machine without problems.

I will deploy it on http://1flow.io/ at next major release.

I patched `simple` & `query`, and didn't find any other obvious locations to patch.

Your feedback is appreciated.

Regards,

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -302,6 +302,26 @@ Combined with ``try ... finally`` it could be used to postpone invalidation:
 Postponing invalidation can considerably speed up batch jobs.
 
 
+Dog-pile effect avoidance
+-------------------------
+
+This package can benefit from `python-redis-lock` to avoid cache dog-pile
+effect (also known as the thundering herd effect or cache stampede). This
+feature is available in QuerySet cache and `@cached_as()`/`@cached_view_as()`
+decorators.
+
+Just define `CACHEOPS_USE_LOCK=True` in settings to make the `@cached`
+decorator protected against this issue.
+
+As it will use another redis call and depends on the creation of a network
+wide lock (that will affect eventual celery workers, etc), it's disabled
+by default.
+
+But the lock is only used on cache misses; there is no consequence on the
+read operations. Overall, this features shouldn't affect performances
+negatively in most cases, and should obviously reduce load on cache expiration.
+
+
 Using memory limit
 ------------------
 
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -3,6 +3,7 @@ import sys
 from functools import wraps
 import json
 import six
+import redis_lock
 from funcy import cached_property, project, once, once_per, monkey
 from funcy.py2 import mapcat, map
 from .cross import pickle, md5
@@ -19,7 +20,8 @@ try:
 except ImportError:
     MAX_GET_RESULTS = None
 
-from .conf import model_profile, redis_client, handle_connection_failure, LRU, ALL_OPS
+from .conf import model_profile, redis_client, handle_connection_failure, \
+                  LRU, ALL_OPS, USE_LOCK
 from .utils import monkey_mix, get_model_name, stamp_fields, load_script, \
                    func_cache_key, cached_view_fab, get_thread_id
 from .tree import dnfs
@@ -89,8 +91,17 @@ def _cached_as(*samples, **kwargs):
             if cache_data is not None:
                 return pickle.loads(cache_data)
 
-            result = func(*args)
-            cache_thing(cache_key, result, cond_dnfs, timeout)
+            if USE_LOCK:
+                with redis_lock.Lock(redis_client, cache_key):
+                    cache_data = redis_client.get(cache_key)
+                    if cache_data is not None:
+                        return pickle.loads(cache_data)
+                    result = func(*args)
+                    cache_thing(cache_key, result, cond_dnfs, timeout)
+            else:
+                result = func(*args)
+                cache_thing(cache_key, result, cond_dnfs, timeout)
+
             return result
 
         return wrapper
@@ -247,10 +258,12 @@ class QuerySetMixin(object):
         # TODO: do not cache empty queries in Django 1.6
         superiter = self._no_monkey.iterator
         cache_this = self._cacheprofile and 'fetch' in self._cacheconf['ops']
+        should_read_cache = not self._cacheconf['write_only'] \
+            and not self._for_write
 
         if cache_this:
             cache_key = self._cache_key()
-            if not self._cacheconf['write_only'] and not self._for_write:
+            if should_read_cache:
                 # Trying get data from cache
                 cache_data = redis_client.get(cache_key)
                 if cache_data is not None:
@@ -261,14 +274,38 @@ class QuerySetMixin(object):
 
         # Cache miss - fallback to overriden implementation
         results = []
-        for obj in superiter(self):
-            if cache_this:
-                results.append(obj)
-            yield obj
 
-        if cache_this:
-            self._cache_results(cache_key, results)
-        raise StopIteration
+        # We use the lock only if we need to cache. If not needing
+        # to cache, using the lock is superfluous anyway because we
+        # *will* dog-pile by computing the queryset multiple times.
+        if cache_this and should_read_cache and USE_LOCK:
+            with redis_lock.Lock(redis_client, cache_key):
+                # Retry fetching from cache in case previous locker filled it.
+                cache_data = redis_client.get(cache_key)
+                if cache_data is not None:
+                    results = pickle.loads(cache_data)
+                    for obj in results:
+                        yield obj
+                    raise StopIteration
+                # implicit else:
+                # We are the first thread / process to acquire the lock,
+                # it's our job to fetch the queryset and feed the cache.
+
+                for obj in superiter(self):
+                    results.append(obj)
+                    yield obj
+                self._cache_results(cache_key, results)
+                raise StopIteration
+        else:
+            # Either we are not caching, or we
+            # are simply not using the lock feature.
+            for obj in superiter(self):
+                if cache_this:
+                    results.append(obj)
+                yield obj
+            if cache_this:
+                self._cache_results(cache_key, results)
+            raise StopIteration
 
     def count(self):
         if self._cacheprofile and 'count' in self._cacheconf['ops']:
--- a/cacheops/simple.py
+++ b/cacheops/simple.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 import os, time
+import redis_lock
+
 from functools import wraps
 from .cross import pickle, md5hex
 
 from django.conf import settings
 
-from .conf import redis_client, handle_connection_failure
+from .conf import redis_client, handle_connection_failure, USE_LOCK
 from .utils import func_cache_key, cached_view_fab
 
 
@@ -34,8 +36,12 @@ class CacheKey(str):
 
 class BaseCache(object):
     """
-    Simple cache with time-based invalidation
+    Simple cache with time-based invalidation and dog-pile effect avoidance.
     """
+
+    # Override this variable in subclasses that implement redis.
+    can_lock = False
+
     def _cached(self, timeout=None, extra=None, _get_key=None):
         """
         A decorator for caching function calls
@@ -47,8 +53,18 @@ class BaseCache(object):
                 try:
                     result = self.get(cache_key)
                 except CacheMiss:
-                    result = func(*args, **kwargs)
-                    self.set(cache_key, result, timeout)
+                    if USE_LOCK and self.can_lock:
+                        # This will block in all calls but the
+                        # first to avoid dog-pile effect.
+                        with redis_lock.Lock(redis_client, cache_key):
+                            try:
+                                result = self.get(cache_key)
+                            except CacheMiss:
+                                result = func(*args, **kwargs)
+                                self.set(cache_key, result, timeout)
+                    else:
+                        result = func(*args, **kwargs)
+                        self.set(cache_key, result, timeout)
 
                 return result
 
@@ -73,6 +89,8 @@ class BaseCache(object):
 
 
 class RedisCache(BaseCache):
+    can_lock = True
+
     def __init__(self, conn):
         self.conn = conn
 
--- a/cacheops/conf.py
+++ b/cacheops/conf.py
@@ -31,7 +31,7 @@ for key in profiles:
 
 LRU = getattr(settings, 'CACHEOPS_LRU', False)
 DEGRADE_ON_FAILURE = getattr(settings, 'CACHEOPS_DEGRADE_ON_FAILURE', False)
-
+USE_LOCK = getattr(settings, 'CACHEOPS_USE_LOCK', False)
 
 # Support DEGRADE_ON_FAILURE
 if DEGRADE_ON_FAILURE:
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ setup(
         'redis>=2.9.1',
         'funcy>=1.2,<2.0',
         'six>=1.4.0',
+        'python-redis-lock==2.0.0',
     ],
 
     classifiers=[
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -92,6 +92,7 @@ else:
     }
 
 CACHEOPS_LRU = bool(os.environ.get('CACHEOPS_LRU'))
+CACHEOPS_USE_LOCK = bool(os.environ.get('CACHEOPS_USE_LOCK'))
 CACHEOPS_DEGRADE_ON_FAILURE = bool(os.environ.get('CACHEOPS_DEGRADE_ON_FAILURE'))
 ALLOWED_HOSTS = ['testserver']
 
--- a/tox.ini
+++ b/tox.ini
@@ -23,6 +23,7 @@ commands =
     ; Test with pre 2.2 style conf
     env CACHEOPS_CONF=old ./run_tests.py []
     env CACHEOPS_LRU=1 ./run_tests.py []
+    env CACHEOPS_USE_LOCK=1 ./run_tests.py []
     env CACHEOPS_DB=postgresql ./run_tests.py []
     ; Djangos pre 1.5 don't work with PostGIS 2.1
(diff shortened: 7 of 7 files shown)


Suor:
Hi, this looks like very solid patch overall. There are some issues with it though:

1. `QuerySet` cache is not protected by locks, see `QuerySetMixin.iterator()`.
2. Unlocking processes should not try to calculate value and recache it, they should at least first try to get it from cache. The way it's implemented now every process will still make the calculation.
3. All processes enter lock sequentially which is not desired behavior if there is a fast path - get value from cache filled by previous process.

Suor:
Thinking a bit more about it I came up with this procedure:
```python
result = None
result_p = cache_get(cache_key)

while result_p is None and result is None:
    lock = Lock(redis_conn, cache_key, expire=1)
    if lock.acquire(blocking=False):
        result = calculate()
        cache_set(cache_key, result)
    else:
        with lock:
            result_p = cache_get(cache_key)

return result or pickle.loads(result_p)
```

Karmak23:
Re,

Thanks for comment 1, I missed it and will update the patch.

About comment 2: yes, my bad. The load is not parallel anymore but now sequential… I will provide a new patch updated from your suggestion ASAP.

About comment 3 and your implementation, yes, my bad too. I missed some part when getting [the python-redis-lock example](https://github.com/ionelmc/python-redis-lock#avoid-dogpile-effect-in-django). However I think the final code will be simpler than yours because what @ionelmc propose seems to give the same result without the `while … acquire(blocking=False)` overhead.

I'm currently leaving for a customer meeting, hoping to send a new patch in a few hours.

Thanks for your review,

Karmak23:
Sorry, my journey didn't go as I wanted. I will work again on `cacheops` tomorrow.

Karmak23:
Hi,

sorry for the delay, here is another patch to complement the first.

Point 1) was tricky, I had to think hard to get to a factorized result, thus the lot of comments to make the implicit cases obviously visible.
Point 2) should be addressed correctly.

I think Point 3) is addressed by my patch too, but I perhaps misunderstood something.

Regards,

README.rst
```diff
@@ -302,6 +302,24 @@ Combined with ``try ... finally`` it could be used to postpone invalidation:
 Postponing invalidation can considerably speed up batch jobs.
 
 
+Dog-pile effect avoidance
+-------------------------
+
+This package can benefit from `python-redis-lock` to avoid cache dog-pile
+effect (also known as the thundering herd effect or cache stampede).
+
+Just define `CACHEOPS_USE_LOCK=True` in settings to make the `@cached`
```
Suor: Please, mention QuerySet cache and `@cached_as()` here too.

cacheops/query.py
```diff
@@ -86,11 +88,23 @@ def wrapper(*args, **kwargs):
             cache_key = 'as:' + _get_key(func, args, kwargs, key_extra)
 
             cache_data = redis_client.get(cache_key)
+
```
Suor: Please, don't add blank lines where they weren't before. This pollutes history and also departs from current terse style.

Also, please remove them with rebase to leave no history of them added and removed.

cacheops/query.py
```diff
@@ -86,11 +88,23 @@ def wrapper(*args, **kwargs):
...
+                    cache_data = redis_client.get(cache_key)
+
+                    if cache_data is not None:
+                        return pickle.loads(cache_data)
+
+                    result = func(*args)
+            else:
+                result = func(*args)
+
             cache_thing(cache_key, result, cond_dnfs, timeout)
```
Suor: Cache write should be in the lock or second process acquiring it will redo the calculation.

cacheops/query.py
```diff
@@ -247,27 +261,72 @@ def iterator(self):
         # TODO: do not cache empty queries in Django 1.6
         superiter = self._no_monkey.iterator
         cache_this = self._cacheprofile and 'fetch' in self._cacheconf['ops']
-
+        cache_for_read = (
```
Suor: Please rename `cache_for_read` to `should_read_cache`. Also, don't break condition into 2 lines, even flakes don't like it :), see Travis logs.

cacheops/query.py
```diff
@@ -247,27 +261,72 @@ def iterator(self):
...
+            not self._cacheconf['write_only']
+            and not self._for_write
+        )
         if cache_this:
             cache_key = self._cache_key()
-            if not self._cacheconf['write_only'] and not self._for_write:
+
+            if cache_for_read:
+
```
Suor: Again, please, no new newlines.

cacheops/query.py
```diff
@@ -247,27 +261,72 @@ def iterator(self):
...
         results = []
-        for obj in superiter(self):
+
+        # We use the lock only if we need to cache. If not needing
+        # to cache, using the lock is superfluous anyway because we
+        # *will* dog-pile by computing the queryset multiple times.
+        if cache_this and cache_for_read and USE_LOCK:
+            with redis_lock.Lock(redis_client, cache_key):
+
+                # In case the lock was aquired by more than one thread or
```
Suor: Lock can't really be acquired by 2 processes, it's its purpose, so message sounds confusing. Probably something like "Retry fetching from cache in case previous locker filled it." will do better.

cacheops/query.py
```diff
@@ -247,27 +261,72 @@ def iterator(self):
...
+                # We are the first thread / process to acquire the lock,
+                # it's our job to fetch the queryset and feed the cache.
+
+                for obj in superiter(self):
+                    results.append(obj)
+                    yield obj
+
+                self._cache_results(cache_key, results)
+
+                # StopIteration is raised at the end of method.
```
Suor: Just raise it and skip comment, it will be more clear.

cacheops/simple.py
```diff
@@ -46,8 +48,20 @@ def wrapper(*args, **kwargs):
                 cache_key = 'c:' + _get_key(func, args, kwargs, extra)
                 try:
                     result = self.get(cache_key)
+
                 except CacheMiss:
-                    result = func(*args, **kwargs)
+                    if USE_LOCK:
+                        # This will block in all calls but the
+                        # first to avoid dog-pile effect.
+                        with redis_lock.Lock(self.conn, cache_key):
```
Suor: Can't really use `self.conn` here cause this is a base class, which is also used for file cache.

cacheops/simple.py
```diff
@@ -46,8 +48,20 @@ def wrapper(*args, **kwargs):
...
+                            try:
+                                result = self.get(cache_key)
+
+                            except CacheMiss:
+                                result = func(*args, **kwargs)
+                    else:
+                        result = func(*args, **kwargs)
+
                     self.set(cache_key, result, timeout)
```
Suor: Again, write should be protected by lock.

Suor:
Overall this looks solid, thanks for all your work so far. I added line notes, please look them through.

I also have some general considerations:

1. You are not testing `CACHEOPS_USE_LOCK=True` code path in any way now. 
    The way you can achieve that is by adding this line to `tests/settings.py`:
```python
CACHEOPS_USE_LOCK = bool(os.environ.get('CACHEOPS_USE_LOCK'))
```
And this to `tox.ini`:
```ini
    env CACHEOPS_USE_LOCK=1 ./run_tests.py []
```

2. Please add `timeout` to all locks, overwise any process acquiring the lock and then failing will cause all others to hang up. Something like 1 second will do, you can also create setting for that, but that's unnecessary.

3. Point 3 is not really taken care of since all locked processes will still go sequentially. But it appears to be impossible to do smarter with redis-lock, you'll need shared/exclusive lock for that. There is no readily available implementation for that and if it were it could have actually made things slower, cause it's more complex. So I think we should stick with current approach for now of unpredictable length :).

cacheops/simple.py
Karmak23: As I use the `cached()` decorator in [1flow](/1flow/1flow) and really need the dog-pile-avoiding feature, I propose the test be `if USE_LOCK and getattr(self, 'conn', None):`. What do you think about it ?

Suor: `getattr(self, 'conn', None)` will work, but it's a dirty hack. Just use `redis_client` directly for locks here for now, I'll think how this could be done better some time in the future.

Karmak23: OK. I've got the rest nearly finished, I will find the time to rebase before the end of week-end I hope. 

About the empty lines, I find the code more readable with, but no problem I will remove them. 

About the other mistakes (cache_set() outside of lock), my current workload & tiredness obviously affects the quality of the code I produce. Sorry for that, but above all, **thanks** for your careful review.

Suor: On code terseness, it's really a matter of habit and a bit of taste. More
empty lines - every code fragment is a bit more clear, less - overall
picture is more evident.

Anyway, thanks very much for all your effort.

Karmak23:
The rebased patch will follow in next minutes.

Only one remains unclear for me: about 2), when you say “*any process acquiring the lock and then failing will cause all others to hang up*”.

What kind of failure are you thinking about ?

If you mean that all processes except the first will block until the first finishes to compute the result, then I think this is exactly intended, because they will be unblocked by a signal at the end of computation, thanks to the original implementation of `python-redis-lock`. 

If they don't block, they are likely to have no result from the cache at the moment they fail to aquire (the cache miss already occured) and they won't compute anything because the computation occurs in the lock. If find this situation unacceptable. They *must* have something to return (or compute it, or wait for it to be computed).

In a high load scenario, if there are a lot of process waiting for the lock while one computes a result, at the end of computation all waiters will all acquire the lock in turn and will the result from inside the lock, and return it immediately. 

In my mind, the "all blocking while one computes and then no fast-path for waiters" is the implicit tradeoff of the feature. This *the* first point of dogpile effect avoiding (the second beiing serving stale data, but it's out of scope of my patch for now). 

At worst, all processes will wait *1 × computation time* + *N × lock/unlock time*, instead of all doing the computation in parallel and hogging the system (in the no-lock-at-all scenario). This is not perfect, but still far better. 

With a timeout, either we need to introduce the `while` you suggested at the first time, which basically introduces some kind of polling (which seems to be what `python-redis-lock` allows to avoid with its back-signal feature), either the method will return no result (= fails), which is not acceptable (cf. above). 

If you mean that a failing first will let the lock in an unknown or blocked state forever (`python-redis-lock` refers to this rare case and provides a management command to clear the locks), then yes, it can happen, but it's very rare. In this case, a timeout would effectively workaround this rare problem. 

But then I strongly need to have long timeout because in most places where I use the cache have more than 10 seconds computations. 

Finally, as this is for a rare bug, I'd prefer to publish the code first as it is, and "see" how much the problem happens (or not) before introducing polling and (relatively) more code complexity. As the lock feature is disabled by default, we can label it as "experimental" and invite users to report issues.

It is still possible I have missed something or not understood what you told, so forgive me if it's the case.

Karmak23:
So here is the rewrapped version with all your remarks. It looks better ;-)
I will let it run for 2-3 days on my development environment to see if the "rare bug" mentionned above occurs or not.

Karmak23:
Only one remains unclear for me: about 2), when you say “any process acquiring the lock and then failing will cause all others to hang up”.

What kind of failure are you thinking about ?

If you mean that all processes except the first will block until the first finishes to compute the result, then I think this is exactly intended. They will be unblocked by a signal at the end of computation, thanks to the original implementation of python-redis-lock.

In a high load scenario where a lot of processes wait for the lock while one computes a result, all waiters will all acquire the lock in turn at the end of computation, and will immediately return the cached result from inside the lock.

In my mind, this workflow is the implicit tradeoff of the locking feature. It's the first point of dogpile effect avoiding (the second beiing serving stale data, but it's out of scope of my patch for now). This is not perfect, but still far better than all of them computing in parallel.

With an eventual timeout to avoid blocking, we need to introduce the `while` you suggested before. This basically introduces polling, which seems to be what python-redis-lock allows to avoid with its back-signal feature.

If you mean that a failing first-process will let the lock in an unknown or blocked state forever (python-redis-lock refers to this rare case and provides a management command to clear the locks), then yes, it can happen, but it's very rare. In this case, a timeout would effectively workaround this problem.

But then, I personnaly strongly need to have a long timeout, because in most places where I use the cache have more than 10 seconds computations (my database is huge and dog-pile is clearly hammering the machine).

As this is for a rare bug, I'd prefer to publish the code first as it is, and "see" how much the problem happens (or not) before introducing polling and (relatively) more code complexity. As the lock feature is disabled by default, we can label it as "experimental" and invite users to report/share issues.

It is still possible I have missed something or not understood what you told, so forgive me if it's the case.

Suor:
By fail I mean crash. And yes, I am worried if block is left in acquired state forever. This effectively means that any process trying to acquire it will hang forever, which will trigger website or whatever completely inoperable until admin comes up and fixes that, and he or she will have a hard time too, cause reboot won't help and tracking an issue to cache could be non-trivial, also if clearing all the cache is not an option admin will need to trace it to certain lock key, which is even harder.

Relying on integrator to write some hacky code using `lock.reset()` is just naive, nobody will do that. And I don't think it's a good approach anyway, timeout is way more reliable.

If 1 second is too small, use 60 or make a setting. This way you won't hang someone site for half a day - a very realistic scenario otherwise.

Karmak23:
That seems a fair worry; automatic recover is likely to be an appreciated feature.

I will update the patch with a dedicated setting and more documentation ASAP.

Karmak23:
Btw, the Travis build failed with a syntax error, I didn't see it. I will create dedicated flake8 config for cacheops, I have too much warnings per file in default config.

Suor:
You can use `tox -e flakes` to run flake8, no need to make your own config.
You can also run tests with:
```bash
pip install -r test_requirements.txt
./runtests.py
CACHEOPS_USE_LOCK=1 ./runtests.py
```

Karmak23:
BTW, there is currently a nasty side-effect when the `redis_client` has a timeout configured:

while the lock is acquired blocking and waiting from release signal, it will fail with `timeout reading from socket`.

I would tend to workaround this problem by creating another client dedicated to locking, with its own timeout value, which should always be greater than the one from in various `@cache*` implementation. This would mean another client in the module.

What's your point of view on this issue ?

Suor:
Sounds sane.

Suor:
Hey, this turned to be harder than it looked initially. Reach me if you need help or have any questions.

Karmak23:
Hi, sorry for the delay, and thanks for your proposition.

I have been very busy last week synching people around working on https://github.com/UnissonCo/dataserver while beiing 1000km away from home ;-)

I'll take time to fix the timeout issue and implement the crash-auto-workaround feature this week. 

This is taking me much time than anticipated, but the handfull of features and reliability is worth it I think.

Karmak23:
Hi @Suor I didn't forget you but have been very busy lately on [python-ftr](/1flow/python-ftr) and building a REST API for [1flow](/1flow/1flow). I'm trying to find time to finish this lock patch. Surely I will make it because I need it, but currently other things have more priority. Sorry for the delay…

ionelmc:
@Karmak23 what's blocking this PR (besides the conflicts)? about the timeout issue, can you explain a bit more - maybe there's something that redis-lock can handle/fix?

Karmak23:
@ionelmc I didn't have the time to fix the timeout bug. I think the patch in its current form will trigger nasty bugs in very loaded conditions. 

We didn't validate it with @Suor because I needed to review it completely

Today I don't have time anymore. Sorry.

Suor:
I think now that python-redis-lock is overkill for this, custom small implementation could be much simpler and more efficient:
- reuse `cache_key` as lock key,
- release all locked threads simultaneously,
- blend release into `cache_thing.lua` script.

Possible lock func:
```python
def get_or_lock(client, key):
    signal_key = key + ':signal'

    while True:
        data = client.get(key)
        if data is None:
            if client.set(key, 'LOCK', nx=True, ex=LOCK_TIMEOUT):
                client.delete(signal_key)
                return None  # Should release lock there and lpush to signal_key
        elif data != 'LOCK':
            return data

        # No data and not locked, wait
        client.brpoplpush(signal_key, signal_key, timeout=LOCK_TIMEOUT)
```
Then add this to cache_thing script:
```lua
redis.call('lpush', signal_key, '1')
redis.call('expire', signal_key, 5)
```
And we are done.

Suor:
BTW, @ionelmc, why did you stop using `EXPIRE` on signal keys?

ionelmc:
@Suor It's still used: https://github.com/ionelmc/python-redis-lock/blob/master/src/redis_lock/__init__.py#L221

Are you referring to something else?

Suor:
I mean you don't rely on expire command to remove old lists.

Suor:
Implemented in 7a89046669473670019aa89fc3d2d16bbb7ddb9a, tests and docs to follow.

Can be used as:
```python
@cached_as(qs, lock=True)
def heavy_func(...):
    # ...
    return res

for item in qs.cache(lock=True):
    # ...
```

I found that global setting would be less useful. It also bypasses locking for cache hits and hence will have zero overhead for them.

ionelmc:
@Suor regarding the expire removal from redis-lock, the reason was that it wasn't reliable, see: https://github.com/ionelmc/python-redis-lock/pull/32 - the issue became apparent when someone tried to replace EXPIRE with PEXPIRE. I suspect on a busy-enough server EXPIRE would have the same problem too. Feel free to give input if you know more about how redis works internally.

Suor:
As far as I understand you only experienced issues when tried to set timeout to some very low value, so leaving it sane like 1 or few seconds should be ok. Also seems far cleaner solution than cleanup thread.

Suor:
Isn't it safe to just make it big enough? A solution with a separate thread
looks much dirtier to me.

ionelmc:
Technically cleanup is not done in thread, only the auto extend is in
thread (completely optional iow)

On Oct 22, 2016 4:51 PM, "Alexander Schepanovski" <notifications@github.com>
wrote:
>
> As far as I understand you only experienced issues when tried to set
timeout to some very low value, so leaving it sane like 1 or few seconds
should be ok. Also seems far cleaner solution than cleanup thread.

Issue #136, opened 2015-03-06
How to invalidate objects in queryset when using bulk update? 


whitefoxx:
No signals were send by django update method, right? so, how to invalidate the objects when using 
```python
model.objects.filter(**kwargs).update(**kwargs)? 
```
I wrote this method to invalidate the objects return by filter method explicitly, will it work?
```python
def nocache_update(qset, **kwargs):
    from cacheops import invalidate_obj
    for obj in qset.iterator():
        try:
            invalidate_obj(obj)
        except:
            pass
    # update must behind invalidate obj
    n = qset.update(**kwargs)
    return n
```
And call this method as follow:
```python
qset = model.objects.filter(**kwargs)
nocache_update(qset, **kwargs)
```

Suor:
This will kind of work, but could be very slow. Also, you should update first, invalidate next, and no need to catch all exceptions:
```python
from cacheops import invalidate_obj

def invalidated_update(qs, **kwargs):
    n = qs.update(**kwargs)
    for obj in qs.iterator():
        invalidate_obj(obj)
    return n
```

If it's too slow you can drop cache for the whole model:
```python
from cacheops import invalidate_model

qs = ...
qs.update(**kwargs)
invalidate_model(qs.model)
```

whitefoxx:
Thank you for your answer. 
It is still a little confusing for me, why update first and invalidate next. In my understanding, the update method will modify even remove some of the objects in qs. If an object was contained in qs before and then removed after update operation, how can I invalidate this object since it's no longer in qs?

Suor:
If you invalidate first and update second you get a race: some other process can come just in the meantime and cache something. If you are worrying about objects get removed from a set you are querying, which is a valid concern if your fields you filter by and fields you update intersect, you can query objects, then make an update and then invalidate objects:
```python
def invalidated_update(qs, **kwargs):
    list(qs) # fill queryset cache
    n = qs.update(**kwargs)
    for obj in qs: # iterate by queryset cache
        invalidate_obj(obj)
    return n
```

rattrayalex:
Can this example be added to the documentation? I think it's a very important one.

Suor:
Handled with `.invalidated_update()` now.

PR #141, opened 2015-04-15
Update() isn't updating the cache at all

EmilStenstrom:
I'm in the process of converting a big Django johnny-cache project to cacheops. I'm unable to get update() queries to work properly. We use them to avoid running signals for some queries.

I've created a failing test case that shows the bug in action.

Would you mind taking a look and see if this is fixable?

### DIFF
--- a/tests/migrations/0002_model141.py
+++ b/tests/migrations/0002_model141.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tests', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Model141',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('is_updated', models.BooleanField(default=False)),
+            ],
+        ),
+    ]
--- a/tests/models.py
+++ b/tests/models.py
@@ -216,3 +216,7 @@ if os.environ.get('CACHEOPS_DB') == 'postgis':
 
     class Geometry(gis_models.Model):
         point = gis_models.PointField(geography=True, dim=3, blank=True, null=True, default=None)
+
+# 141
+class Model141(models.Model):
+    is_updated = models.BooleanField()
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -480,6 +480,10 @@ class IssueTests(BaseTestCase):
         with self.assertNumQueries(0):
             list(All.objects.cache(ops='all').all())
 
+    def test_141(self):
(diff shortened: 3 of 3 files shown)


Suor:
This is by design, see CAVEATS, 4.

Suor:
If you select single object as in test, then you can call `invalidate_obj()` for that. Or just use `.save()` on that not mass `.update()` call. If you do actual mass update, then you can `invalidate_model()`, which will clear all cache for a model. You can also look at #136.

I don't think there is reasonable fits everyone behaviour, which is also efficient fir this case, so cacheops does nothing.

Suor:
Also, for you case disconnecting or switching off with a flag signals you don't want would probably be better solution. This way you bypass all the signals, including ones that trigger cacheops (and god only knows what else).

EmilStenstrom:
I see, I should have read the CAVEATES more carefully.

The rationale for purposefully deactivating signals is that this is how we have implemented denormalized fields in this project. When a Model is updated, we have attached several post_save signals to that model. Each signal saves down denormalized information on the model objects, that is later used to avoid SQL joins queries. Most of our denormalized fields work like this, and sprinkling them with invalidate_obj seems like a very dirty solution.

Did johnny-cache implement invalidation for update queries very inefficiently?

Suor:
johnny-cache always invalidates at table level so it doesn't have this issue.

I didn't understand the problem with denormalized fields and `invalidate_obj()`. What do you mean by sprinkling them? `invalidate_obj()` doesn't send any signals it only invalidates cache related to object, has nothing to do with denormalized fields.

I still think using ordinary `obj.save()` will work better for you. You can use `update_fields` kwarg to only update desired fields, `update_fields` is passed to `post_save` signals so you can introspect it there to only recalculate if certain fields are updated:

```python
@receiver(post_save, sender=YourModel)
def recalc_f(sender, instance, **kwargs)
    updated_fields = kwargs.get('updated_fields')
    if updated_fields and 'g' not in updated_fields:
        return

    instance.f = ...
```

EmilStenstrom:
Just to give some more feedback on this: I've managed to convert a big Django codebase to cacheops. The biggest surprise to me was that updates weren't invalidated at all. After checking each instance of update I found no reason to worry about invalidation performance, I could simply do a loop with invalidate_obj and be done with it. This is what I meant by "sprinkling", I now have a cacheops import everywhere where I use update(), something this feels a bit dirty.

But overall, cacheops seems to work great, and I want to expend a thanks from our team to you! Thanks!

EmilStenstrom:
Another minor thing: When writing a test for this I saw that the instructions didn't mention that you need to create a migration for your model. This took me while to figure out so I suggest that you add that to the "How to write a test" docs.

Suor:
That sucks. Tests project used to be without migrations, but it's not possible since Django 1.8. What is worse that automatically generated migrations won't work cross python.

I'll try modifying test suite to generate migrations on the fly.

PR #145, opened 2015-05-04
Cacheops signals invalidate an object after all other post_save() signals

EmilStenstrom:
Cacheops automatically invalidates instances with a post_save signal. That signal is added after all existing post_save signals. This means that we will get stale data if we do queries inside other signals.

This commit adds a test that does a simple get() query inside a post_save signal, and fails because it only gets access to stale data, not the data that should be saved to the database (this is post_save after all).

### DIFF
--- a/tests/migrations/0002_one.py
+++ b/tests/migrations/0002_one.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tests', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='One',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('boolean', models.BooleanField(default=False)),
+            ],
+        ),
+    ]
--- a/tests/models.py
+++ b/tests/models.py
@@ -217,3 +217,17 @@ if os.environ.get('CACHEOPS_DB') == 'postgis':
 
     class Geometry(gis_models.Model):
         point = gis_models.PointField(geography=True, dim=3, blank=True, null=True, default=None)
+
+# 145
+class One(models.Model):
+    boolean = models.BooleanField(default=False)
+
+def set_boolean_true(sender, instance, created, **kwargs):
+    if created:
+        return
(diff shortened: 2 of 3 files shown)


Suor:
Your test fails cause you create an object with `boolean=False`. Note that `Model.objects.create()` saves object, to create object without saving it just use `obj = One()`.

Also, cache is off for `One` model you are using.

Suor:
Also, why you fetch an object in `post_save`? There is an `instance` argument with that object.

EmilStenstrom:
Thanks for looking at the test, and sorry for being sloppy. 

I've now updated the test so that it doesn't run the signal on create (signal returns if created is True), and enabled caching for the One model.

This is how I've tested that it works:
FAIL: ```./run_tests.py IssueTests.test_145```
SUCCESS: ```CACHEOPS_DEGRADE_ON_FAILURE=True ./run_tests.py IssueTests.test_145``` (with redis-server turned off)

I also have a fix for this issue that simply attaches cacheops signals before all other signals, and enabling that patch makes the test pass. I'll commit it if you think this is a valid issue.

EmilStenstrom:
About the rationale for doing queries inside signals:
It's a simplified testcase for the issue we chatted about before, something that works without the cache, and stops working when you enable it.

The real case has a Dialog model with a post_save signal. That signal updates a DashboardItem model that has a foreignkey to the Dialog and that reads data from the Dialog and saves it on the DashboardItem. It's a way for denormalizing data so that we don't have to query Dialogs directly, and can just do filters on the DashboardItems.

All code that recalculates the denormalized field lives inside the DashboardItem, and since it has a foreignkey to the dialog, it just accesses that foreignkey to get the dialog. That foreignkey access is the equivalent of the One.objects.get() fetch in the test. The signal could be rewritten to pass instance to the DashboardItem, but in other cases we might not have a dialog instance directly when recalulating, so using the foreignkey seems better.

I could go through the project and ensure that every signal uses the instance instead of refetching it. But I would rather make sure that cacheops did what I expected, and invalidated the cache directly after .save(), but before ALL post_save signals.

Suor:
Hmm, I see this is the issue for newer djangos. If you have a fix then I'll appreciate a pull request.

EmilStenstrom:
Would it be possible to issue a new release of cacheops for this? Would love to use the something from pypi instead of referencing github.

Suor:
I'll wait a few days to see if we missed something and then make point release.

PR #147, opened 2015-05-05
Make sure cacheops signals are attached first.

EmilStenstrom:
This fix makes sure the object is invalidated before all custom post_save signals are triggered.

It does this by directly manipulating a private Django API; the "receivers" attribute on the signals. This could be dangerous, but I only rely on it being a list, so this should work even over quite big changes to the API.

Fixes #145

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -330,6 +330,11 @@ class QuerySetMixin(object):
             invalidate_obj(obj)
         return objs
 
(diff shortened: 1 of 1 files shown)


Suor:
Should work. The only issue with this patch is adding `.connect_first()` to manager. Make it a simple function please.

EmilStenstrom:
Updated to be a separate method, not on the manager.

PR #151, opened 2015-05-27
Add docs/example for CacheopsLibrary.decorator_tag

frankier:
Firstly, thanks for the project. It's feels really smooth to be able to write simply declare which querysets invalidate something rather than manually peppering signal handlers throughout the source code or having to set really short timeouts. 

With Django's templates it's a common situation to not be to not be dealing directly in querysets, but also in functions returning querysets and templatetags that make their own querysets. It's also probably reasonably common for a template fragment to depend upon more than one queryset. In addition, cache invalidation and cache key naming fall outside what's typically thought of as template/view type stuff.

Therefore the current {% cached_as %} and {% cached %} are not sufficient for many template fragment caching needs. In the project I'm currently working on our current approach is to copy paste the whole of the templatetags.py to customise template fragment caching behaviour in python. This patch exposes make_cached_tag and make_cached_as_tag to help people make their own cached_as-like tags, incorporating whatever application code they need to to contruct the invalidation querysets and cache key. 

TODO:
 * Could the interface be better? (comments solicited)
 * Bring back invalidate_fragment (hints welcome)
 * Documentation

I'm posting this to get some feedback before doing any more work on this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -464,6 +464,74 @@ To invalidate cached fragment use:
 
     invalidate_fragment(fragment_name, extra1, ...)
 
+If you have more complex fragment caching needs, cacheops provides a helper to
+make your own templatetags which decorate a template fragment in a way
+analagous to decorating a function using ``@cached`` or ``@cached_as``.
+
+First ``cacheops.templatestags.CacheopsLibrary`` is used in place of
+``django.template.Library``. Then ``register.decorator_tag`` turns a function
+``foor`` returning a decorator into a ``{% foo %}`` ``{% endfoo %}`` pair.
+``register.decorator_tag`` passes through arguments to the templatetag to your
+function and optionally takes an argument, ``takes_context`` in the same way as
+Django's own ``inclusion_tag`` and ``simple_tag``.
+
+To motivate this functionality, here's example of how two navigation menus at
+the top and bottom of a page might be cached. In this example the cache
+fragment:
+
+1. is invalidated on multiple querysets/models: the menu items and a feature
+   flag
+2. varies on seleted item
+3. varies on and current language.
+
+In addition the cache is bypassed entirely for staff users.
+
+In `myapp/templatetagss/mycachetags.py`:
+
+.. code:: python
+
+   from cacheops import cached_as
+   from cacheops.templatetags import CacheopsLibrary
+
+   register = CacheopsLibrary()
+
+   def dont_cache(func):
+       return func
+
+   @register.decorator_tag(takes_context=True)
+   def cache_menu(context, menu_name):
+       from django.utils import translation
+       from featureflag.models import Flag
+       from myapp.models import MyMenuItem
+
+       request = context.get('request')
+       if request and request.user.is_staff():
+           return dont_cache
+
+       return cached_as(
+           Flag.objects.filter(name='menu_feature'),
+           MyMenuItem,
+           timeout=24 * 60 * 60,
+           extra=(
+               "menu",
+               menu_name
+               translation.get_language(),
+               context.get('mymenuitem_selected')))
+
+It could then be used in templates as follows:
+
+.. code:: django
+
+    {% load mycachetags %}
+
+    {% cache_menu "top" %}
+        ... the top menu template code ...
+    {% endcache_menu %}
+    ... some template code ..
+    {% cache_menu "bottom" %}
+        ... the bottom menu template code ...
+    {% endcache_menu %}
+
 
 Jinja2 extension
 ----------------


frankier:
Here's a usage example to show what I'm talking about. Say I wanted to cache a whole menu which:
 * contained Mezzanine pages
 * an special item only when a flag is enabled
 * could be in multiple languages
 * varied in content depending on the current page
 * was editable by staff

Here's how I might do it with the new functions:

    def cache_menu(context):
        from django.utils import translation
        from mezzanine.pages.models import Page
        from waffle.models import Flag
        from myapp.models import MyMenuItem

        request = context.get('request')
        user = request.user if request else None
        return (
            # avoid Mezzanine's CurrentSiteRelatedManager hitting Site
            QuerySet(Page).all(),
            Flag.objects.filter(name='menu_feature'),
            MyMenuItem,
        ), dict(
            timeout=24 * 60 * 60,
            extra=(
                "menu",
                translation.get_language(),
                context.get('page'),
                context.get('mymenuitem_selected'),
                user.pk if user.is_staff else None
            )
        )

    register.tag(*make_cached_as_tag(cache_menu))

Some feedback about whether you'd be willing to merge a more complete version of this would be appreciated.

Additional TODO:
 * Make user's function return a single dict with a samples key instead of returning an args, kwargs tuple.
 * Allow user's function to return False to disable caching altogether (in the above example, we might instead return False when user.is_staff is True)

Suor:
I see the use and I appreciate your efforts to contribute back, thanks. I think this could be done. There some issues:
- `invalidate_fragment()` should stay
- your code is messy (which makes sense since your started from already messy one ^)

I think we can resolve second one, which is serious 'cause it's already hard for me, and I guess you too, to add features there. The solution is probably classes here, we are using inversion of control anyway:
```python
class CachedFragment(TagHelperNode):
    def __init__(self, parser, token):
        # ...
        self.args, self.kwargs = parse_bits(...)

    def render(self, context):
        def _render():
            return carefully_strip_whitespace(nodelist.render(context))

        args, kwargs = self.get_resolved_arguments(context)
        decorator = self.get_decorator(args, kwargs)
        return decorator(_render)()

    def get_decorator(...):
        raise NotImplementedError

    #...


class cached_as(CachedFragment):
    def get_decorator(self, resolved_args, resolved_kwargs):
        return cached_as(*resolved_args, **resolved_kwargs)
```

Or something like this, it's just a sketch.

Suor:
Or we could still use decorators, only decorate this `.get_decorator()` behaviour.

```python
@tag_from_decorator
def cached_as(queryset, timeout, fragment_name, *extra):
    return cacheops.cached_as(queryset, timeout=timeout, extra=(fragment_name,) + extra)
```

This looks best, we need to alter decorator and arguments and this is exactly what we have here. I should have implemented it that way from the start.

Suor:
This should help - 99cf2fa91c437724391618fd9efb13703f7526bc.

frankier:
Yep, agreed doing it that was around is a lot neater. I made some changes to export decorator_tag and added some docs. Doing it this way makes it easy to bypass the cache too so I mostly just added docs.

Suor:
I don't want to make this public functionality just yet. So I won't merge it for now. 

There are few other issues with your code:
- why are you changing from decorator to call syntax?
- why are you separating `decorator_tag` into `register.tag + decorator.tag`? this goes against django practice.
- your example, while being nice, won't work 'cause `context` is not passed. Maybe I need to add `takes_context` argument to `@decorator_tag`.

Suor:
Added optional `takes_context` argument to `@decorator_tag` in https://github.com/Suor/django-cacheops/commit/31b31e30fd63ec3d518635ef98f44c581a046375.

frankier:
>why are you changing from decorator to call syntax?

Related to the second question, we need to save the original function for invalidate_tag. Otherwise we'd overwrite the original with the compile function.

>why are you separating decorator_tag into register.tag + decorator.tag? this goes against django practice.

As I understand it the current way would cause all tags to be registered in the cacheops tag namespace which seems a bit odd/quirky. The alternative is to inherit from django.template.Library and add decorator_tag as a method which seems like a reasonable alternative. If you think this better I can rewrite + rebase it this way.

>your example won't work

Oops. Yeah well spotted.

Suor:
> As I understand it the current way would cause all tags to be registered in the cacheops tag namespace which seems a bit odd/quirky. 

Yeah, it looks weird. But this way I don't need to create a bulky dir structure `app/templatetags/app.py` and can just add a decorator tag next to views.

Suor:
Ok, I changed that to library approach in fe3c937e95a66821b6be593c9437080afa7da0d3.
So you can do:

```python
from cacheops import CacheopsLibrary

register = CacheopsLibrary()

@register.decorator_tag
def lalala(...):
    # ...

# or

@register.decorator_tag(takes_context=True)
def lalala(context, ...):
    # ...
```

Suor:
I should probably publish it now, given all the efforts. Like experimental feature.

Can you update your docs/code sample?

frankier:
Works great! Updated to just add docs now (hopefully a little clearer too).

Issue #82, opened 2014-04-10
AttributeError: Cause 'NoneType' object has no attribute '_meta'


gotlium:
Stacktrace:

```
  File "django/core/handlers/base.py", line 111, in get_response
    response = callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 494, in wrapper
    return wrapped(*args, **kwargs)
  File "django/contrib/auth/decorators.py", line 20, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "apps/app/views.py", line 582, in apps_data
    data['apps'] = app.count()
  File "cacheops/query.py", line 370, in count
    return cached_as(self, extra='count')(self._no_monkey.count)(self)
  File "cacheops/query.py", line 102, in wrapper
    result = func(*args)
  File "django/db/models/query.py", line 351, in count
    return self.query.get_count(using=self.db)
  File "django/db/models/sql/query.py", line 398, in get_count
    obj = self.clone()
  File "django/db/models/sql/query.py", line 259, in clone
    obj.where = copy.deepcopy(self.where, memo=memo)
  File "python2.7/copy.py", line 174, in deepcopy
    y = copier(memo)
  File "django/utils/tree.py", line 61, in __deepcopy__
    obj.children = copy.deepcopy(self.children, memodict)
  File "python2.7/copy.py", line 163, in deepcopy
    y = copier(x, memo)
  File "python2.7/copy.py", line 230, in _deepcopy_list
    y.append(deepcopy(a, memo))
  File "python2.7/copy.py", line 174, in deepcopy
    y = copier(memo)
  File "django/utils/tree.py", line 61, in __deepcopy__
    obj.children = copy.deepcopy(self.children, memodict)
  File "python2.7/copy.py", line 163, in deepcopy
    y = copier(x, memo)
  File "python2.7/copy.py", line 230, in _deepcopy_list
    y.append(deepcopy(a, memo))
  File "python2.7/copy.py", line 163, in deepcopy
    y = copier(x, memo)
  File "python2.7/copy.py", line 237, in _deepcopy_tuple
    y.append(deepcopy(a, memo))
  File "python2.7/copy.py", line 174, in deepcopy
    y = copier(memo)
  File "django/db/models/query.py", line 52, in __deepcopy__
    obj = self.__class__()
  File "cacheops/query.py", line 221, in __init__
    self._cacheprofile = model_profile(self.model)
[...]

PR #154, opened 2015-06-16
Fix crash caused by `None` being passed as a model.

samdoiron:
This error [seems to have occurred before](https://github.com/Suor/django-cacheops/issues/82) and may have recently gone into regression.

This change appears to fix the problem for my use case, and the git history implies that this is expected behaviour.


Suor:
`model_profile()` should not be patched to silently mask error as you propose. The real issue causing `None` to be passed there should be fixed. 

So how do you get this?

PR #163, opened 2015-09-09
Proxy models caching issue

bourivouh:
Hi!
Proxy and concrete models have the same cache_key. It's okay, cause they have the same data in db. But if concrete model is cached and you trying to fetch proxy model, you will get concrete model instance anyway. And vise versa: if proxy model is cached, you will get proxy model instance even if you trying to get concrete model.
This is brief example:
```
In [1]: Video.objects.cache().get(pk=1)
Out [1]: <Video: Video object>
In [2]: VideoProxy.objects.cache().get(pk=1)
Out [2]: <Video: Video object>
In [3] VideoProxy.objects.nocache().get(pk=1)
Out [3] <VideoProxy: VideoProxy object>
```

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -816,6 +816,20 @@ class ProxyTests(BaseTestCase):
         with self.assertRaises(NonCachedMedia.DoesNotExist):
             MediaProxy.objects.cache().get(title=media.title)
 
+    def test_proxy_caching(self):
+        video = Video.objects.create(title='Pulp Fiction')
+        self.assertEqual(type(Video.objects.cache().get(pk=video.pk)),
+                         Video)
+        self.assertEqual(type(VideoProxy.objects.cache().get(pk=video.pk)),
+                         VideoProxy)
+
+    def test_proxy_caching_reversed(self):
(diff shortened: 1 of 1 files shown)


Suor:
Thanks for the tests. 

There is obvious way to fix this - add model class to cache key, this will, however, lead to caching same data twice. A smarter way to fix this would be to cache once and replace class on fetch, which could be unsafe if __init__ or some signal is writing something to one class, but not another.

To decide in design with more context I would like to hear about your use case. Why are you using proxy models and why are you querying both base and descendant ones?

bourivouh:
The major usecase is like this.
We have User model. This is customized django.contrib.auth.User model. A lot of models in another apps relate to it. For example, User is an author of Video.
UserChannel is a proxy model for User. It doesn't have diffrent fields yet, so it's just a proxy. But it has own properties and methods, which cannot be in User model because of semantic reasons and cross imports.
So if user register or authorize, he deal with base model. If editors add his channel to another catalogs or feeds, they deal with proxy models.

Suor:
Fixed the obvious way.

PR #164, opened 2015-10-08
Add support to cache_on_save for multiple field get

dvincelli:
One of the apps I work on commonly uses a two field lookup so it would be nice to maintain the cache by  these fields as the key. This PR adds that ability but note that I am still in the process of evaluating the change (I am sharing it early to hear your feedback). I also added coverage for the existing cache_on_save option of specifying a single field for the key. 

NB: I elected not to use a dictionary comprehension (dict([(k,v) for...]) vs {k: v for ...} to support older versions of python.

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -412,13 +412,19 @@ class ManagerMixin(object):
             for k in unwanted_dict:
                 del instance.__dict__[k]
 
-            key = 'pk' if cache_on_save is True else cache_on_save
+            if cache_on_save is True:
+                keys = ['pk']
+            elif isinstance(cache_on_save, six.string_types):
+                keys = [cache_on_save]
+            else:
+                keys = cache_on_save
             # Django doesn't allow filters like related_id = 1337.
             # So we just hacky strip _id from end of a key
             # TODO: make it right, _meta.get_field() should help
-            filter_key = key[:-3] if key.endswith('_id') else key
-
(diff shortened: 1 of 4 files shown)


Suor:
First issue with this is ambigous API: when you write `{'cache_on_save': ['name', 'title']}` it's easy to presume that it will precache both
```python
Model.objects.get(name=...)
Model.objects.get(title=...)
```
And not
```python
Model.objects.get(name=..., title=...)
```

Also, caching still happens once you do a get, however complicated it is, `cache_on_save` is merely an optimization. So the question is how hard do you need this? Are updating more frequently than reading?

dvincelli:
@suor I agree with you that the API I propose is ambiguous.

I have a code path that is very hot both for reads and writes but I believe I can effectively cache the read and conditionally issue the write. This means I may not need the write-through. However, I consistently observe selects to the database. I am still digging into my framework to understand where some of the selects are coming from but I noticed that there is a mix of query by pk and my field combination.

On the other hand, there is a possibility that I may need this type of cache behaviour to get the desired performance.

Finally, it appears that I have ordering issues related to kwargs which is causing cache misses (this is a guess as to why I have some randomly failing tests).

But you are correct, I may not need this feature at all.

dvincelli:
I solved my issues in another way. I'll revisit if it the need arises again. Thanks for you input @Suor

Issue #22, opened 2012-11-28
Race condition when in transaction


Suor:
Data can be invalidated on post_save and recached before actual commit hits the database.

rattrayalex:
How can this be solved?

Suor:
On application level this could be solved by manual invalidation after commit. Adding appropriate `invalidate_obj()` or such. On cacheops level - by monkey patching transaction handling routines.

rattrayalex:
Cool. This seems like a standard caveat - it probably makes sense not to have cacheops solve it for people. But it would be really nice to have this in the docs =)

rattrayalex:
I'd be happy to submit this w/ related changes to readme if you're interested.

Suor:
Cacheops can solve this, I just never really needed this features and overall demand is low. The lack of interest is the primary reason this is not addressed yet.

Suor:
This is finally fixed, thanks to @jhillacre.

PR #165, opened 2015-10-15
Use a local cache during atomic transactions

jhillacre:
In these commits, I've added support to cacheops for django's  `transaction.atomic` context manager. My co-workers and myself have run into situations where rolled back data was making it into the cache. I noticed this was on the readme's todo list, so I hope these changes are something you like for cacheops.

In this new functionality, a thread specific local cache is setup on  `atomic`. Keys that are usually stored in redis are instead held in a local cache . Since `atomic` relates to database transactions, this functionality only affects querysets and views.

When `atomic` commits, it moves local keys to redis and then clears the local cache. When `atomic`  rolls back, it clears the local cache. Inside `atomic` blocks, cacheops will check the local cache before redis. Local cache misses cause cacheops to look to redis as before these changes.

This new functionality also deals with savepoints. Upon entering a nested `atomic` block, a local cache context is setup. When `atomic` commits a savepoint, cacheops moves keys up to the parent's context. This parent can be another save point or the outer transaction. When `atomic` rolls back a savepoint, cacheops will discard the context.

To support these different contexts, the local cache is a [`collections.ChainMap`](https://docs.python.org/3/library/collections.html#chainmap-objects) instance. Python 3.3 introduced `collections.ChainMap`. A [backport is available](https://pypi.python.org/pypi/chainmap) for earlier versions.  I have added this backport as a cacheops package dependency.

I have added a setting for this functionality called `CACHEOPS_RESPECT_ATOMIC`, which defaults to false. Because of the dependency on `transaction.Atomic`, cacheops ignores this setting unless running django 1.6+.

I would like to hear any feedback you have on this pull request. Some of my specific questions are:

* Do you think `CACHEOPS_RESPECT_ATOMIC` should be true by default instead of false?

* My co-workers and myself are not using the Simple time-invalidated cache or File Cache. My understanding that these are not related to database transactions. With this assumption, I did not add transaction local caching code to these functions. Is this correct?

* While developing these changes, I was running the full `run_tests.py` suite. After passing these tests, I moved the functionality behind the  `CACHEOPS_RESPECT_ATOMIC` setting. I wrote some tests of my own, with the `@override_settings` decorator to turn this setting on. I also have the `BasicTests` being rerun with this setting on.

  * Do you think more of the previous tests should be rerun with the new setting on?

  * Would you review the tests I have added? I'm not sure if they can be more simple or organized better.

* Should I rebase the commits in this pull request to a single commit?

### DIFF
--- a/cacheops/conf.py
+++ b/cacheops/conf.py
@@ -1,9 +1,6 @@
 # -*- coding: utf-8 -*-
-import warnings
-from cacheops.redis_client import LocalCachedTransactionRedis
 import six
-import redis
-from funcy import memoize, decorator, identity, merge
+from funcy import memoize, merge
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
@@ -13,47 +10,10 @@ ALL_OPS = {'get', 'fetch', 'count', 'exists'}
 LRU = getattr(settings, 'CACHEOPS_LRU', False)
 DEGRADE_ON_FAILURE = getattr(settings, 'CACHEOPS_DEGRADE_ON_FAILURE', False)
 
-
-# Support DEGRADE_ON_FAILURE
-if DEGRADE_ON_FAILURE:
-    @decorator
-    def handle_connection_failure(call):
-        try:
-            return call()
-        except redis.ConnectionError as e:
-            warnings.warn("The cacheops cache is unreachable! Error: %s" % e, RuntimeWarning)
-        except redis.TimeoutError as e:
-            warnings.warn("The cacheops cache timed out! Error: %s" % e, RuntimeWarning)
-else:
-    handle_connection_failure = identity
-
-class SafeRedis(LocalCachedTransactionRedis):
-    get = handle_connection_failure(LocalCachedTransactionRedis.get)
-
-
-class LazyRedis(object):
-    def _setup(self):
-        # Connecting to redis
-        try:
-            redis_conf = settings.CACHEOPS_REDIS
-        except AttributeError:
-            raise ImproperlyConfigured('You must specify CACHEOPS_REDIS setting to use cacheops')
-
-        client = (SafeRedis if DEGRADE_ON_FAILURE else LocalCachedTransactionRedis)(**redis_conf)
-
-        object.__setattr__(self, '__class__', client.__class__)
-        object.__setattr__(self, '__dict__', client.__dict__)
-
-    def __getattr__(self, name):
-        self._setup()
-        return getattr(self, name)
-
-    def __setattr__(self, name, value):
-        self._setup()
-        return setattr(self, name, value)
-
-redis_client = LazyRedis()
-
+try:
+    REDIS_CONF = settings.CACHEOPS_REDIS
+except AttributeError:
+    raise ImproperlyConfigured('You must specify CACHEOPS_REDIS setting to use cacheops')
 
 @memoize
 def prepare_profiles():
--- a/cacheops/conf.py
+++ b/cacheops/conf.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 import warnings
-from cacheops.redis import LocalCachedTransactionRedis
+from cacheops.redis_client import LocalCachedTransactionRedis
 import six
 import redis
 from funcy import memoize, decorator, identity, merge
--- a/cacheops/conf.py
+++ b/cacheops/conf.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import warnings
+from cacheops.redis import LocalCachedTransactionRedis
 import six
 import redis
 from funcy import memoize, decorator, identity, merge
@@ -26,8 +27,8 @@ if DEGRADE_ON_FAILURE:
 else:
     handle_connection_failure = identity
 
-class SafeRedis(redis.StrictRedis):
-    get = handle_connection_failure(redis.StrictRedis.get)
+class SafeRedis(LocalCachedTransactionRedis):
+    get = handle_connection_failure(LocalCachedTransactionRedis.get)
 
 
 class LazyRedis(object):
@@ -38,7 +39,7 @@ class LazyRedis(object):
         except AttributeError:
             raise ImproperlyConfigured('You must specify CACHEOPS_REDIS setting to use cacheops')
 
-        client = (SafeRedis if DEGRADE_ON_FAILURE else redis.StrictRedis)(**redis_conf)
+        client = (SafeRedis if DEGRADE_ON_FAILURE else LocalCachedTransactionRedis)(**redis_conf)
 
         object.__setattr__(self, '__class__', client.__class__)
         object.__setattr__(self, '__dict__', client.__dict__)
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import sys
-import json
 import threading
 from cacheops import CacheMiss
 import six
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -45,7 +45,7 @@ def cache_thing(cache_key, data, cond_dnfs, timeout):
     try:
         # are we in a transaction?
         Atomic.thread_local.cacheops_transaction_cache[cache_key] = {
-            'data': data,
+            'data': pickle.dumps(data, -1),
             'cond_dnfs': cond_dnfs,
             'timeout': timeout,
             # these two help us out later for possible invalidation
@@ -64,8 +64,6 @@ def cache_thing(cache_key, data, cond_dnfs, timeout):
             timeout
         ]
     )
-
-_marker = object()
 
 def cached_as(*samples, **kwargs):
     """
@@ -105,16 +103,6 @@ def cached_as(*samples, **kwargs):
         @wraps(func)
         def wrapper(*args, **kwargs):
             cache_key = 'as:' + key_func(func, args, kwargs, key_extra)
-
-            # try transaction local cache first
-            try:
-                cache_data = Atomic.thread_local.cacheops_transaction_cache.get(cache_key, None)
-            except AttributeError:
-                # not in transaction
-                pass
-            else:
-                if cache_data is not None and cache_data.get('data', _marker) is not _marker:
-                    return cache_data['data']
 
             cache_data = redis_client.get(cache_key)
             if cache_data is not None:
@@ -292,22 +280,9 @@ class QuerySetMixin(object):
                 # Trying get data from cache
                 results = None
 
-                # try transaction local cache first
-                try:
-                    cache_data = Atomic.thread_local.cacheops_transaction_cache.get(
-                        cache_key, None
-                    )
-                except AttributeError:
-                    # not in transaction
-                    pass
-                else:
-                    if cache_data is not None and cache_data.get('data', None) is not None:
-                        results = cache_data['data']
-
-                if results is None:
-                    cache_data = redis_client.get(cache_key)
-                    if cache_data is not None:
-                        results = pickle.loads(cache_data)
+                cache_data = redis_client.get(cache_key)
+                if cache_data is not None:
+                    results = pickle.loads(cache_data)
 
                 if results is not None:
                     for obj in results:
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -3,14 +3,11 @@ import sys
 import json
 import threading
 import six
+from threading import local
 from funcy import select_keys, cached_property, once, once_per, monkey, wraps
 from funcy.py2 import mapcat, map
 from .cross import pickle, md5
-try:
-    from .transaction import Atomic
-except ImportError:
-    # django is too old for this
-    Atomic = None
+from .transaction import AtomicMixIn
 
 import django
 from django.utils.encoding import smart_str
@@ -25,6 +22,11 @@ try:
     from django.db.models.query import MAX_GET_RESULTS
 except ImportError:
     MAX_GET_RESULTS = None
+try:
+    from django.db.transaction import Atomic
+except ImportError:
+    Atomic = None
+
 
 from .conf import model_profile, redis_client, handle_connection_failure, LRU, ALL_OPS
 from .utils import monkey_mix, stamp_fields, load_script, \
@@ -532,6 +534,11 @@ def install_cacheops():
     monkey_mix(QuerySet, QuerySetMixin)
     QuerySet._cacheprofile = QuerySetMixin._cacheprofile
     QuerySet._cloning = QuerySetMixin._cloning
+
+    # django.db.transaction.Atomic exists in Django 1.6+
+    if Atomic:
+        monkey_mix(Atomic, AtomicMixIn)
+        Atomic.thread_local = local()
 
     # DateQuerySet existed in Django 1.7 and earlier
     # Values*QuerySet existed in Django 1.8 and earlier
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -22,10 +22,7 @@ try:
     from django.db.models.query import MAX_GET_RESULTS
 except ImportError:
     MAX_GET_RESULTS = None
-try:
-    from django.db.transaction import Atomic
-except ImportError:
-    Atomic = None
+from django.db.transaction import Atomic
 
 
 from .conf import model_profile, redis_client, handle_connection_failure, LRU, ALL_OPS
@@ -45,21 +42,20 @@ def cache_thing(cache_key, data, cond_dnfs, timeout):
     """
     Writes data to cache and creates appropriate invalidators.
     """
-    if Atomic:
-        try:
-            # are we in a transaction?
-            Atomic.thread_local.cacheops_transaction_cache[cache_key] = {
-                'data': data,
-                'cond_dnfs': cond_dnfs,
-                'timeout': timeout,
-                # these two help us out later for possible invalidation
-                'db_tables': [x for x, y in cond_dnfs],
-                'cond_dicts': [dict(i) for x, y in cond_dnfs for i in y]
-            }
-            return
-        except AttributeError:
-            # we are not in a transaction.
-            pass
+    try:
+        # are we in a transaction?
+        Atomic.thread_local.cacheops_transaction_cache[cache_key] = {
+            'data': data,
+            'cond_dnfs': cond_dnfs,
+            'timeout': timeout,
+            # these two help us out later for possible invalidation
+            'db_tables': [x for x, y in cond_dnfs],
+            'cond_dicts': [dict(i) for x, y in cond_dnfs for i in y]
+        }
+        return
+    except AttributeError:
+        # we are not in a transaction.
+        pass
     load_script('cache_thing', LRU)(
         keys=[cache_key],
         args=[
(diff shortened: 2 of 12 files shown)


jhillacre:
There is something weird happening in the mysql tests that I think I need another pair of eyes on.

In https://travis-ci.org/Suor/django-cacheops/jobs/85635763 `test_expressions_save (tests.tests.TransactionalBasicTests)` fails. I can also reproduce this on my machine.

When I run those tests on their own, I don't get the failed test.
```bash
(venv)[joel@asper django-cacheops] (transactions)$ env CACHEOPS_DB=mysql ./run_tests.py TransactionalBasicTests
Creating test database for alias 'default'...
Creating test database for alias 'slave'...
...............
----------------------------------------------------------------------
Ran 15 tests in 0.082s

OK
Destroying test database for alias 'default'...
Destroying test database for alias 'slave'...
```

cacheops/query.py
```diff
@@ -85,6 +105,17 @@ def decorator(func):
...
 
+            # try transaction local cache first
+            if Atomic:
+                try:
+                    cache_data = Atomic.thread_local.cacheops_transaction_cache.get(cache_key, None)
+                except AttributeError:
+                    # not in transaction
+                    pass
+                else:
+                    if cache_data is not None and cache_data.get('data', None) is not None:
```
Suor: `cached_data['data']` could legitimately be `None`.

Suor:
First, this is a big piece of work, so thank you for tackling this. [The issue](https://github.com/Suor/django-cacheops/issues/22) was sitting there for too long.

And now some considerations:

1. I was planning to go to cacheops 3.0 to drop older Djangos support there. Looks like your code will also benefit from that, less ifs and catching ImportErrors. Also, we can go with transactions support enabled by default this way. So no `CACHEOPS_RESPECT_ATOMIC` setting at all.
   
    I will update code to drop older Djangos support and so you can rebase onto it. What version of Django do you use?

2. You are using `cacheops.transaction.atomic()` in tests, which is wrong - cacheops should work with code using `django.db.transaction.atomic()` because this is what people will use. Or third-party code people are using will use, as well as Django itself. To support this monkey patching original Atomic will be most probably required, see how `QuerySet` is patched.
   
   Also tests should not access `Atomic.thread_local....`,  only public behavior should be checked. 

3. Now in `cache_thing()` you both write to transaction cache and to redis. This could lead to concurrent process reading uncommitted data from redis. You do the same in `invalidate_dict()`. I propose using extending `redis_client` so that when in transaction all write requests to it would be queued and then either applied or discarded on commit or rollback. Read requests should be smart to refer to queue first and then go to redis on miss. Essentially the same thing you are doing now everywhere, but at redis client level.

jhillacre:
1. >  I was planning to go to cacheops 3.0 to drop older Djangos support there. Looks like your code will also benefit from that, less ifs and catching ImportErrors. Also, we can go with transactions support enabled by default this way. So no CACHEOPS_RESPECT_ATOMIC setting at all.

  > I will update code to drop older Djangos support and so you can rebase onto it. What version of Django do you use?

  I've been using djanjo 1.8.5 / python 2.7 / postgres 9.3 to stay with our production instances. I hadn't really tried other django versions / py versions / db software until I completely failed all the travis-ci tests earlier.

2.  >  You are using cacheops.transaction.atomic() in tests, which is wrong - cacheops should work with code using django.db.transaction.atomic() because this is what people will use. Or third-party code people are using will use, as well as Django itself.

  I'll switch the tests to import `django.db.transaction` instead.

  > To support this monkey patching original Atomic will be most probably required, see how QuerySet is patched.

  I had been monkey patching `django.db.transaction.atomic`, but not in a consistent way to what you were already doing. I'll change it my class to a mixin for use by `monkey_mix` in `install_cacheops`.

  > Also tests should not access Atomic.thread_local...., only public behavior should be checked.

  I wasn't sure how to write a test for local vs remote cache otherwise, since the public API is agnostic to which cache. Should the transaction specific tests be cut down to just checking if savepoint & transaction rollbacks drop cached items properly?

3. > Now in cache_thing() you both write to transaction cache and to redis. This could lead to concurrent process reading uncommitted data from redis. You do the same in invalidate_dict(). I propose using extending redis_client so that when in transaction all write requests to it would be queued and then either applied or discarded on commit or rollback. Read requests should be smart to refer to queue first and then go to redis on miss. Essentially the same thing you are doing now everywhere, but at redis client level.

  In the op, I mentioned I wasn't sure if it was expected for this transactional handling to apply to 'Simple time-invalidated cache' or 'File Cache' APIs. If it does, then moving this code as you specified seems like a good plan.

jhillacre:
Your line comment regarding `cache_data['data']` is allowed to be None should be addressed in 87b7899.

Suor:
> I had been monkey patching django.db.transaction.atomic, but not in a consistent way to what you were already doing. I'll change it my class to a mixin for use by monkey_mix in install_cacheops.

The way you monkey patch now would fail if someone does:
 ```python
from django.db.transaction import atomic
```
before your patching. This is why I am patching methods.

> I wasn't sure how to write a test for local vs remote cache otherwise, since the public API is agnostic to which cache. Should the transaction specific tests be cut down to just checking if savepoint & transaction rollbacks drop cached items properly?

Several things to check:
- that things are not in a cache until final commit. The way to do this using public API is using or simulating second thread,
- that rolled back things are not written into cache and not accessible within same transaction after rollback,
- maybe something else, basically composed cache + db should behave as db only: read only commited things in other threads, read all written things except rolled back inside same thread.

> In the op, I mentioned I wasn't sure if it was expected for this transactional handling to apply to 'Simple time-invalidated cache' or 'File Cache' APIs. If it does, then moving this code as you specified seems like a good plan.

This has nothing to do with simple or file cache. This is about not mixing all sort of transation code into `cache_thing()`, `@cached_as()`, `QuerySet.iterator()`, `invalidate_dict()` and such. This how I envision this:
- make `cacheops/redis.py`,
- subclass `StrictRedis` there and use that to make `redis_client` to use everywhere,
- override clients `.get()`, add `.run_script()` to replace current `load_script()()`, make this things respect transactions and savepoints.

This will allow to leave all the cache writing, fetching and invalidating code as is.

**Comment by Suor** [comment]
Dropped older Djangos in master branch, you can now rebase and assume Django 1.7+ everywhere in your code.

**Comment by jhillacre** [comment]
> Dropped older Djangos in master branch, you can now rebase and assume Django 1.7+ everywhere in your code.

The PR branch is now rebased.

> > I wasn't sure how to write a test for local vs remote cache otherwise, since the public API is agnostic to which cache. Should the transaction specific tests be cut down to just checking if savepoint & transaction rollbacks drop cached items properly?

> Several things to check:
> * that things are not in a cache until final commit. The way to do this using public API is using or simulating second thread,
> * that rolled back things are not written into cache and not accessible within same transaction after rollback,
> * maybe something else, basically composed cache + db should behave as db only: read only commited things in other threads, read all written things except rolled back inside same thread.

I've got the tests using only public APIs now, using real threads to check the cache from outside transactions/savepoints. I don't think I quite understood your last point, but please see if the tests I got cover what you were thinking of.

There was an issue getting the other threads to work with django's default of using in-memory sqlite databases for tests. The workaround I used is unix specific. See the changes I made in tests/settings.py for details. If unix specific tests are an issue, we can make django use a on disk sqlite database for tests by copying `['NAME']` to `['TEST']['NAME']` in `DATABASES`.

File: cacheops/conf.py [hunk 2]
```diff
@@ -38,7 +39,7 @@ def _setup(self):
         except AttributeError:
             raise ImproperlyConfigured('You must specify CACHEOPS_REDIS setting to use cacheops')
 
-        client = (SafeRedis if DEGRADE_ON_FAILURE else redis.StrictRedis)(**redis_conf)
+        client = (SafeRedis if DEGRADE_ON_FAILURE else LocalCachedTransactionRedis)(**redis_conf)
```
Suor: If `DEGRADE_ON_FAILURE` is `True` then transactions are not supported.

cacheops/query.py
```diff
@@ -257,9 +274,13 @@ def iterator(self):
             cache_key = self._cache_key()
             if not self._cacheconf['write_only'] and not self._for_write:
                 # Trying get data from cache
+                results = None
+
                 cache_data = redis_client.get(cache_key)
                 if cache_data is not None:
                     results = pickle.loads(cache_data)
+
+                if results is not None:
```
Suor: Shouldn't need a check here. If something returned from `redis_client.get()` is not `None` then it should work.

cacheops/conf.py
Suor: Probably move all redis class to `cacheops/redis.py` and import from where. `conf.py` will become data only, which is good.

tests/tests.py
```diff
@@ -938,3 +939,228 @@ def test_invalidate_model_with_geometry(self):
...
+        super(ThreadWithReturnValue, self).join(*args, **kwargs)
+        return self._return
+
+
+def get_category_filter_pk_1():
+    try:
+        return list(Category.objects.filter(pk=1).cache())
+    finally:
+        from django.db import connection
+        connection.close()
```
Suor: Why explicit `.close()`?

tests/tests.py
```diff
@@ -938,3 +939,228 @@ def test_invalidate_model_with_geometry(self):
...
+        t.start()
+        committed_remote_cache_results = t.join()
+        self.assertEqual(
+            committed_remote_cache_results[0].title,
+            'Fjango'
+        )
+
+    def test_savepoint_with_rollback(self):
+        with atomic():
+            try:
```
Suor: There should be changes in outer block too to test this fully.

Suor:
Bigger things:

1. Local cache structure knowledge shouldn't be used in various places. It should be collected in single place either in `cacheops/transaction.py` or `cacheops/redis.py`. Probably the last one (cause it needs it to implement `.get()`. anyway) with `Atomic` just calling `redis_client.begin()`, `.commit()` and `.rollback()`.
2. It would be nice to queue invalidation as well when in transaction. With the exception of `invalidate_all()`, which should probably stay completely oblivious of transactions. I don't think it should even flush local cache.
3. Also would be nice to make tests check that cache is used by counting number of queries with `self.assertNumQueries()`. 

Smaller ones in line comments.

Suor:
To achieve 1 you will probably need to move `cache_thing` and `invalidate` script calls into `redis_client.cache_thing()` and `redis_client.invalidate()` methods.

Suor:
Actually queueing invalidation is not nice to have it's essential. Few scenarious where you currently have race:

1. Cache then rollback:
    - start transaction,
    - update a post,
    - fetch the post via title (misses local cache, hits db and caches new state),
    - rollback,
    - fetch the post via title, got rolled back state from cache.
2. Cache by second thread before commit:
    - start transaction,
    - update a post (redis cache is invalidated),
    - fetch the post in second thread (misses cache, hits db and caches old state),
    - commit,
    - old post state in cache.

The issue here is no invalidation on commit/rollback. But we also need to invalidate immediately cause this thread could fetch and miss local, but hit redis cache if it queries the same object diffrently. Invalidating both places also won't work:

- start transaction,
- update post,
- fetch the post via title in second thread (caches old state),
- fetch the post via title in current thread (fetches old state from cache),

So the only solution I see is queueing invalidation until commit and disabling caching while in transaction at all or at least for modified tables. Queueing `cache_thing` is unnecessary, it's a mere optimization.

tests/tests.py
jhillacre: I was getting connections left open sometimes in Postgres when the test went to delete the CB. Manual closing was recommended in a django issue. I'll see if I can re find the link and put it in the source.

Joel

cacheops/conf.py
jhillacre: > If DEGRADE_ON_FAILURE is True then transactions are not supported.

Picking this up after a little less then a week. My understanding of DEGRADE_ON_FAILURE is that `True` means turn redis.ConnectionError and redis.TimeoutError into warnings. `False` would mean these exceptions bubble to the top. Am missing some knowledge on why DEGRADE_ON_FAILURE's value would effect transactions?

Suor: This line means that `LocalCachedTransactionRedis` is not used if `DEGRADE_ON_FAILURE` is on, which will lead to `redis_client.get()` never use local cache and some your logic failing. Your tests will probably fail now with `DEGRADE_ON_FAILURE = True`.

Anyway, rather than focusing on smaller things like this, it could be better to solve the main problem first - that local cache can't solve transactions. Only queueing invalidations will do. You can however reuse your chainmap and things for that.

cacheops/redis_client.py
```diff
@@ -0,0 +1,215 @@
...
+        try:
+            # are we in a transaction?
+            self._local.cacheops_transaction_cache[cache_key] = {
+                'data': data,
+                'cond_dnfs': cond_dnfs,
+                'timeout': timeout,
+                # these two help us out later for possible invalidation
+                'db_tables': [x for x, y in cond_dnfs],
+                'cond_dicts': [dict(i) for x, y in cond_dnfs for i in y]
+            }
```
Suor: As local cache is not really needed to support transactions, I suggest removing it for now. It could be added later with separately. This way we will hopefully do less bugs.

Also, there should be better code to check if we are in transaction, like:
```python
if self.in_transaction:
    ...
else:
    ...
```
Should be trivial to add property like that.

cacheops/redis_client.py
```diff
@@ -0,0 +1,215 @@
...
+                            if match or not cond_dict:
+                                # deep delete, to deal with savepoints in the cache
+                                for mapping in local_cache.maps:
+                                    if key in mapping:
+                                        del mapping[key]
+
+        self.load_script('invalidate')(args=[
+            db_table,
+            json.dumps(obj_dict, default=str)
+        ])
```
Suor: Should not be called right away if in transaction, but queued:
```python
self._local.queue.append((self.load_script('invalidate'), dict(args=[...])))
```
And then run the queue on commit.

cacheops/redis_client.py
```diff
@@ -0,0 +1,215 @@
...
+                if db_table in value.get('db_tables', []):
+                    # deep delete, to deal with savepoints in the cache
+                    for mapping in local_cache.maps:
+                        if key in mapping:
+                            del mapping[key]
+
+        conjs_keys = redis_client.keys('conj:%s:*' % db_table)
+        if conjs_keys:
+            cache_keys = redis_client.sunion(conjs_keys)
+            redis_client.delete(*(list(cache_keys) + conjs_keys))
```
Suor: Same as `invalidate_dict()`, should be queued when in transaction.

cacheops/redis_client.py
```diff
@@ -0,0 +1,215 @@
...
+        except AttributeError:
+            # not in transaction
+            pass
+        else:
+            if cache_data is not None and cache_data.get('data', _marker) is not _marker:
+                return cache_data['data']
+        cache_data = super(LocalCachedTransactionRedis, self).get(name)
+        if cache_data is None:
+            raise CacheMiss
+        return pickle.loads(cache_data)
```
Suor: `.get()` should not break inherited contract: return binary data (not unpickled) and never raise `CacheMiss`. You can add separate method for this behaviour.

tests/tests.py
```diff
@@ -938,3 +939,301 @@ def test_invalidate_model_with_geometry(self):
...
+        new_object[0]._cacheprofile['cache_on_save'] = True
+        new_object[0].title = title
+        new_object[0].save()
+
+    def assert_local_cache_title(self, title):
+        local_cache_results = list(Category.objects.filter(pk=1).cache())
+        self.assertEqual(
+            local_cache_results[0].title,
+            title
+        )
```
Suor: How do you know it's taken from local cache here and not redis or db?

jhillacre:
Sorry to be gone for so long. Replying to your comments below.

> Also would be nice to make tests check that cache is used by counting number of queries with self.assertNumQueries().

I've improved the tests I wrote in regard to this.

> Actually queueing invalidation is not nice to have it's essential.

Agreed.

> So the only solution I see is queueing invalidation until commit and disabling caching while in transaction at all or at least for modified tables.

By using a local cache, I believe that cacheops can handle cache data transitionally. Please check out the latest diffs and let me know if I'm missing out on something here.

> Queueing `cache_thing` is unnecessary, it's a mere optimization.

I don't agree, I think queueing of both cache and invalidation is required. Redis having cached data that was not committed is what alerted me to the need for transaction handling in cacheops. One of my clients was having data that was attempted to be saved on an admin form, shown on a separate PDF report later, even though the save had been rolled back due to an exception after save.

> How do you know it's taken from local cache here and not redis or db?

I've changed get api now so it's possible to specify to get only from the local cache. I don't think anyone but the tests will need it.

> `.get()` should not break inherited contract: return binary data (not unpickled) and never raise CacheMiss. You can add separate method for this behaviour.

I was thinking the local cache would not need to be pickled, but ran into mutation issues with rolled back savepoints. I have now moved pickling back out of the redis client. Callers of cache_thing now need to do pickling of the data to be stored. `.get()` no longer throws CacheMiss.

jhillacre:
I'm going to start looking into the travis ci test failures in Django 1.7.

The failure happening in both Django 1.7 & 1.8, `test_weird_custom (tests.tests.WeirdTests)`, has me confused. I moved the `pickle.dumps(data, -1)` from inside `cache_thing` to outside `redis_client.cache_thing`, but I'm not sure why that would cause the pickle to fail.

jhillacre:
I should also mention that I have rebased up to your newest commit on master, 7336ffd

jhillacre:
> Also, there should be better code to check if we are in transaction, like:

> ```python
if self.in_transaction:
    ...
else:
    ...
```

> Should be trivial to add property like that.

This is implemented now. `in_transaction` is thread specific so it lives at `self._local.in_transaction`.

**Comment by Suor** [comment]
> By using a local cache, I believe that cacheops can handle cache data transitionally. Please check out the latest diffs and let me know if I'm missing out on something here.

Just turn off caching while in transaction. You will need to turn off cache reads anyway 'cause redis has stale data outside transaction and local cache can't possibly fix that. So we end up with queueing writes, which of questionable value and will also lead to races - if 2 transactions change single object concurrently by updating different fields then last finishing one might overwrite the effect of the first one.

The initial issue you were having is transaction mishandling not performance, so local cache might not really be needed at all.

By trying to make both local cache and transactions in single run here were already screwed up logic at least twice. So we should stop, remove any local cache logic from this pull request and concentrate on making transactions right.

**Comment by jhillacre** [comment]
I had been thinking of local cache as the method for transaction management. If we are not queueing writes, does this describe what your looking for instead?

* cache writes happen live to redis.
* before a write, read the value and hold onto it.
* on roll back of any savepoint / transaction, restore cached values to redis.
* invalidations deletes happens live to reids
* before invalidation delete, read the value and hold onto it.
* on roll back of any savepoint / transaction, restore invalidated values to redis.

**Comment by Suor** [comment]
> I had been thinking of local cache as the method for transaction management. 

This won't work. Local cache can't be used for transaction management. It could only be used as optimization, but it still should be throwed away on transaction end (both on success and failure). Writing it to redis will cause data races.

> * cache writes happen live to redis.
> * before a write, read the value and hold onto it.
> * on roll back of any savepoint / transaction, restore cached values to redis.

No redis cache writes nor reads while in transaction. No local cache, do not hold anything. Nothing to write upon commit.

> * invalidations deletes happens live to redis
> * before invalidation delete, read the value and hold onto it.
> * on roll back of any savepoint / transaction, restore invalidated values to redis.

Invalidations happening in a transaction are queued, applied on commit, thrown away on rollback. When invalidation is queued it should not be also executed right away, just queued.

**Comment by jhillacre** [comment]
> Invalidations happening in a transaction are queued, applied on commit, thrown away on rollback. When invalidation is queued it should not be also executed right away, just queued.

Ok, I think I understand.

Does it still make sense to do this as a redis client class? With `redis_client.get()` and `cache_thing()` remaining untouched, I am thinking it would make sense for me to start over from master and have any changes concentrated on `invalidation.py`.

**Comment by Suor** [comment]
Starting over makes sense, start a new pull request and call this research :).

Here are my thoughts on implementation. It would be nice to start from implementing:
```python
def is_in_transaction():
    ...

# to be used as
```

Suor:
Starting over makes sense, start a new pull request and call this one research :).

Here are my thoughts on implementation. It would be nice to start from implementing:
```python
def in_transaction():
    ...

# to be used as
cache_this = ... and not in_transaction()
```

And:
```python
def queue_when_in_transaction(func):
    ...

# to be used as
@queue_when_in_transaction
@handle_connection_failure
def invalidate_dict(...):
    ...
```
And when just add commit/rollback logic. All these should go to `cacheops/transactions.py` as well as `Atomic` patching. 

Moving all redis client stuff to `cacheops/redis.py` still makes sense, but it's unrelated to transactions now, I'll just move it myself as a cleanup.

Suor:
Done moving redis client stuff.

PR #166, opened 2015-10-16
Post lookup signals to collect information about cache hits and misses

paksu:
This pull request adds a signal that users can listen to and collect statistics about `django-cacheops` lookup hit rate.

`cache_read` signal is emitted every time `django-cacheops` hits or misses cache when doing `.get(..)` or `.filter(..)` or `@cached_as` decorator makes a cache lookup.

For example if one wanted to collect data about cache hit rates for individual models with `statsd` they could do this in their Django app:

```
from statsd.defaults.django import statsd
from cacheops import signals

def stats_collector(sender, func, hit, **kwargs):
    signal_origin = func.__name__.lower()
    statsd_path = 'cacheops.hit.%s' % signal_origin if hit else 'cacheops.miss.%s' % signal_origin
    statsd.incr(statsd_path)

signals.cache_read.connect(stats_collector)
```

Emitting a signal with no connected listeners shouldn't degrade performance so it would up to the user to decide how they want to process this data. 

Because Django signals are synchronous doing expensive operations in a signal listener would degrade performance.

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -27,6 +27,7 @@ from .redis import redis_client, handle_connection_failure, load_script
 from .tree import dnfs
 from .invalidation import invalidate_obj, invalidate_dict, no_invalidation
 from .transaction import in_transaction
+from .signals import post_lookup
 
 
 __all__ = ('cached_as', 'cached_view_as', 'install_cacheops')
@@ -251,6 +252,10 @@ class QuerySetMixin(object):
             clone._cloning = self._cloning - 1 if self._cloning else 0
             return clone
 
+    def _send_post_lookup_signal(self, hit_cache):
+        model_path = "%s.%s" % (self.model.__module__, self.model.__name__)
+        post_lookup.send(sender='cacheops.iterator', model=model_path, hit_cache=hit_cache)
+
     def iterator(self):
         # TODO: do not cache empty queries?
         superiter = self._no_monkey.iterator
@@ -267,6 +272,8 @@ class QuerySetMixin(object):
                 if cache_data is not None:
                     results = pickle.loads(cache_data)
                     for obj in results:
+                        # Notify about cache hit
+                        self._send_post_lookup_signal(hit_cache=True)
                         yield obj
                     raise StopIteration
 
@@ -279,6 +286,8 @@ class QuerySetMixin(object):
 
         if cache_this:
             self._cache_results(cache_key, results)
+        # Notify about not hitting cache
+        self._send_post_lookup_signal(hit_cache=False)
         raise StopIteration
 
     def count(self):
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -27,7 +27,7 @@ from .redis import redis_client, handle_connection_failure, load_script
 from .tree import dnfs
 from .invalidation import invalidate_obj, invalidate_dict, no_invalidation
 from .transaction import in_transaction
-from .signals import post_lookup
+from .signals import cache_read
 
 
 __all__ = ('cached_as', 'cached_view_as', 'install_cacheops')
@@ -93,6 +93,7 @@ def cached_as(*samples, **kwargs):
             cache_key = 'as:' + key_func(func, args, kwargs, key_extra)
 
             cache_data = redis_client.get(cache_key)
+            cache_read.send(sender=None, func=func, hit=cache_data is not None)
             if cache_data is not None:
                 return pickle.loads(cache_data)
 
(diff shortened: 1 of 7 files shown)


cacheops/query.py
```diff
@@ -261,6 +265,8 @@ def iterator(self):
                 if cache_data is not None:
                     results = pickle.loads(cache_data)
                     for obj in results:
+                        # Notify about cache hit
+                        self._send_post_lookup_signal(hit_cache=True)
```
Suor: Would be called for each item in a queryset. If fetch is empty, but still a cache hit, then it won't be called.

cacheops/query.py
```diff
@@ -248,6 +248,10 @@ def clone(self, klass=None, setup=False, **kwargs):
             clone._cloning = self._cloning - 1 if self._cloning else 0
             return clone
 
+    def _send_post_lookup_signal(self, hit_cache):
+        model_path = "%s.%s" % (self.model.__module__, self.model.__name__)
+        post_lookup.send(sender='cacheops.iterator', model=model_path, hit_cache=hit_cache)
```
Suor: Useless sender. It will be always the same for this signal class.

Also, why stringify `self.model`? This is against Django way.

Suor:
Would be really nice to see benches, both without and with hander attached. Also, `@cached_as()`  adn simple cache should send signals too if we already started this.

To fix tests you need to add `mock` to `tox.ini -> [testenv] -> deps`.

cacheops/query.py
paksu: I wanted to provide some context about where the signal was emitted from. Maybe passing  `self.model` here as sender would be better.

paksu:
I will do some benchmarking and take a look at `cached_as` signalling. Will post benchmarks here.

cacheops/query.py
paksu: Oops. I will fix this and add a test that ensures signal will be emitted only once per queryset evaluation

paksu:
I added signaling for `cached_as` and benchmarks for signals.

Benchmarks ran on my machine:
```
pickle             time: 0.05ms
unpickle           time: 0.04ms
get_no_cache       time: 0.34ms
get_no_cache_with_signal time: 0.35ms
get_hit            time: 0.43ms
get_hit_with_signal time: 0.44ms
get_miss           time: 0.84ms
get_miss_with_signal time: 0.84ms
count_no_cache     time: 0.23ms
count_hit          time: 0.28ms
count_miss         time: 0.27ms
fetch_construct    time: 0.01ms
fetch_no_cache     time: 0.22ms
fetch_hit          time: 0.34ms
fetch_miss         time: 0.75ms
fetch_cache_key    time: 0.06ms
filter_cache_key   time: 0.09ms
common_construct   time: 0.24ms
common_inplace     time: 0.14ms
common_cache_key   time: 0.21ms
invalidate_obj     time: 0.18ms
save_obj           time: 1.09ms
complex_construct  time: 0.49ms
complex_inplace    time: 0.40ms
complex_cache_key  time: 0.25ms
big_invalidate     time: 0.36ms
model_invalidate   time: 0.85ms
```

cacheops/signals.py
```diff
@@ -0,0 +1,4 @@
+import django.dispatch
+
+post_queryset = django.dispatch.Signal(providing_args=["hit_cache"])
+post_cached_as = django.dispatch.Signal(providing_args=["function_name", "hit_cache"])
```
Suor: Just use a single signal name it `cache_read` with `function` and `hit` arguments.

cacheops/query.py
```diff
@@ -81,6 +82,9 @@ def _get_queryset(sample):
     if not timeout:
         timeout = min(qs._cacheconf['timeout'] for qs in querysets)
 
+    def _send_post_cached_as_signal(cached_func, hit_cache):
+        post_cached_as.send(sender=None, function_name=cached_func.__name__, hit_cache=hit_cache)
+
```
Suor: Just inline this.

cacheops/query.py
```diff
@@ -88,10 +92,12 @@ def wrapper(*args, **kwargs):
 
             cache_data = redis_client.get(cache_key)
             if cache_data is not None:
```
Suor: One line here:
```python
cache_read.send(sender=None, func=func, hit=cache_data is not None)
```
Current behaviour of delaying miss events is inconsistent.

cacheops/query.py
```diff
@@ -248,6 +254,9 @@ def clone(self, klass=None, setup=False, **kwargs):
             clone._cloning = self._cloning - 1 if self._cloning else 0
             return clone
 
+    def _send_post_queryset_signal(self, hit_cache):
+        post_queryset.send(sender=self.model, hit_cache=hit_cache)
+
```
Suor: Also inline this.

Suor:
Sorry for slow response, it slipped through.

This is not benches that I meant. The thing I am intereseted in is the cost for the majority of people not using this. So bench current cacheops against your branch with no handlers attached.

Also, see API/code points in line notes.

paksu:
Okay, I will create benchmarks this week :+1:

Suor:
Would be nice to have it in 3.0, which I am heading to.

Suor:
Merged and cleaned up, thanks.

PR #171, opened 2015-10-29
Adding transactional handling to cacheops using `django.db.transaction.atomic()`

jhillacre:
Based on the conversation in #165, this pull requests adds transactional support to cacheops. Features are:
* While in a transaction, invalidation methods are delayed until the current transaction is completed.
* Delayed invalidation methods that are in a savepoint that is rolled back are dropped.
* Delayed invalidation methods that are in a transaction that is rolled back are dropped.
* While in a transaction, reads are only done live. The data in redis won't know what is happening within the transaction.
* While in a transaction, writes are ignored. This prevents data that is rolled back from making it to redis.


cacheops/invalidation.py
```diff
@@ -33,11 +36,13 @@ def invalidate_obj(obj):
     model = non_proxy(obj.__class__)
     invalidate_dict(model, get_obj_dict(model, obj))
 
+
```
Suor: Please don't add empty lines where they weren't before.

cacheops/query.py
```diff
@@ -85,14 +88,20 @@ def _get_queryset(sample):
...
-                return pickle.loads(cache_data)
+                cache_data = redis_client.get(cache_key)
+                if cache_data is not None:
+                    return pickle.loads(cache_data)
 
             result = func(*args, **kwargs)
-            cache_thing(cache_key, result, cond_dnfs, timeout)
+            # ignore writes while in transaction. no point in writing data that might be rolled back
+            if use_cache:
+                cache_thing(cache_key, result, cond_dnfs, timeout)
```
Suor: Just 
```python
if in_transaction():
    return func(*args, **kwargs)
```
at the start o the wrapper, no need to change the rest of the code.

cacheops/query.py
```diff
@@ -38,14 +39,16 @@ def cache_thing(cache_key, data, cond_dnfs, timeout):
...
+    # ignore writes while in transaction. no point in writing data that might be rolled back.
+    if not in_transaction():
+        load_script('cache_thing', CACHEOPS_LRU)(
+            keys=[cache_key],
+            args=[
+                pickle.dumps(data, -1),
+                json.dumps(cond_dnfs, default=str),
+                timeout
+            ]
+        )
```
Suor: Should be used higher level like in `.iterator()`, `@cached_as()` and cache_on_save code. Just assert here.

cacheops/transaction.py
```diff
@@ -0,0 +1,185 @@
...
+    returns all non list items in a list of lists of of arbitrary depth at a single depth,
+     ignoring non list items that are `None`.
+    """
+    for item in items:
+        if isinstance(item, list):
+            # todo: use yield from in python 3
+            for x in _flatten_and_ignore_none(item):
+                yield x
+        elif item is not None:
+            yield item
```
Suor:
```python
from funcy import iflatten, ikeep, is_list

ikeep(iflatten(items, follow=is_list))
```

cacheops/transaction.py
```diff
@@ -0,0 +1,185 @@
...
+     considered contexts. the latest context is the deepest list who's right most boundary is also
+     the right most boundary of the outer most context. `None` items are added to the list, in order
+     to make second latest contexts the new latest context (when an inner most context ends). These
+     `None` objects are not return on __iter__.
+
+     You probably want to sub class this class to override commit_transaction, or in alternate cases
+      rollback_transaction.
+    """
+    def __init__(self):
+        self._queue = None
```
Suor: Should start from `[]`

cacheops/transaction.py
```diff
@@ -0,0 +1,185 @@
...
+        return self._queue is not None
+
+    def append(self, item):
+        latest_context = _find_latest_context(self._queue)
+        if latest_context is None:
+            latest_context = self._queue
+        latest_context.append(item)
+
+    def __iter__(self):
+        return _flatten_and_ignore_none(self._queue)
```
Suor: Simplify this class interface to `.begin()`, `.commit()` and `.rollback()`. Nested begins are savepoints. Each level has a list in `.queue`, so:
```python
assert queue._queue = []

queue.begin()
assert queue._queue = [[]]

queue.append(call1)
queue.append(call2)
queue.begin()
queue.append(call3)
queue.begin()

assert queue._queue = [[call1, call2], [call3], []]
```
On rollback we just drop last item, on commit we concat last item to second last and then drop last one:
```python
self._queue[-2] += self.pop()
```
When we commit highest level we call `.apply()`:
```python
for func, args, kwargs in self._queue[0]:
    func(*args, **kwargs)
```

cacheops/transaction.py
```diff
@@ -0,0 +1,185 @@
...
+            'func': <some callable>
+            'args': [],
+            'kwargs': {}
+        }
+    these items are executed in order at transaction commit time.
+    """
+    def commit_transaction(self):
+        for item in self:
+            item['func'](*item['args'], **item['kwargs'])
+        super(FunctionQueue, self).commit_transaction()
```
Suor: Use tuples `(func, args, kwargs)`. Also there is no need for separate class.

tests/tests.py
```diff
@@ -13,14 +17,24 @@
...
 from .models import *
 
 
-class BaseTestCase(TestCase):
+# django.test.TestCase wraps individual tests in atomic blocks. Old tests were not written to check
+#  the change in functionality cacheop's transaction support implements when in atomic blocks.
+#  Changing to django.test.TransactionTestCase fixes this issue. Tests are not in atomic blocks.
+#  This does make tests run slower then before, since tables are truncated and fixtures reapplied
+#  between tests.
+class BaseTestCase(TransactionTestCase):
```
Suor: Still unclear why all the tests should be wrapped into `TransactionTestCase`. This seems unnecessary.

tests/tests.py
```diff
@@ -13,14 +17,24 @@
...
     def setUp(self):
         super(BaseTestCase, self).setUp()
         invalidate_all()
+        # for mysql, django will add the query 'SET SQL_AUTO_IS_NULL = 0' before the first query
+        #  run a query before our tests in order to trigger this before we start counting queries.
+        Category.objects.nocache().exists()
```
Suor: Unclear why this is needed, it wasn't needed before.

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+            self.assertTrue(in_transaction())
+        self.assertFalse(in_transaction())
+        with self.assertNumQueries(0):
+            self.assertEqual('Django', list(Category.objects.filter(pk=1).cache())[0].title)
+        self.assertEqual(
+            'Django',
+            return_from_other_thread(
+                lambda: list(Category.objects.filter(pk=1).cache())[0].title
+            )
+        )
```
Suor: Please move all transactions tests to separate file like [here](https://github.com/Suor/django-cacheops/compare/transactions-tests).

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+        executed = len(self)
+        if executed != self.num:
+            raise AssertionError(
+                "%d queries executed, %d expected\nCaptured queries were:\n%s" % (
+                    executed, self.num,
+                    '\n'.join(
+                        query['sql'] for query in self.captured_queries
+                    )
+                )
+            )
```
Suor: Why this is needed?

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+
+
+class TransactionalInvalidationTests(BaseTestCase):
+    fixtures = ['basic']
+
+    def test_atomic_block_change_and_invalidate_obj(self):
+        with self.assertNumQueries(1):
+            self.assertEqual('Django', list(Category.objects.filter(pk=1).cache())[0].title)
+        with self.assertNumQueries(0):
+            self.assertEqual('Django', list(Category.objects.filter(pk=1).cache())[0].title)
```
Suor: No need to check that cache works.

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+            self.assertEqual(
+                'Django',
+                return_from_other_thread(
+                    lambda: list(Category.objects.filter(pk=1).cache())[0].title
+                )
+            )
+            obj = list(Category.objects.filter(pk=1).cache())[0]
+            obj.title = 'Changed'
+            obj.save()
+            invalidate_obj(obj)
```
Suor: Should not call `invalidate_obj()` explicitly here.

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+
+    def test_atomic_block_change_and_invalidate_obj(self):
+        with self.assertNumQueries(1):
+            self.assertEqual('Django', list(Category.objects.filter(pk=1).cache())[0].title)
+        with self.assertNumQueries(0):
+            self.assertEqual('Django', list(Category.objects.filter(pk=1).cache())[0].title)
+        with atomic():
+            self.assertTrue(in_transaction())
+            with self.assertNumQueries(1):
+                self.assertEqual('Django', list(Category.objects.filter(pk=1).cache())[0].title)
```
Suor: No need to check for number of queries, this tests should only check logic now.

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+class IntentionalRollback(Exception):
+    pass
+
+
+class TransactionalInvalidationTests(BaseTestCase):
+    fixtures = ['basic']
+
+    def test_atomic_block_change_and_invalidate_obj(self):
+        with self.assertNumQueries(1):
+            self.assertEqual('Django', list(Category.objects.filter(pk=1).cache())[0].title)
```
Suor: Please use `Category.objects.get(pk=1)` instead of list.

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+            'Changed',
+            return_from_other_thread(
+                target=lambda: list(Category.objects.filter(pk=1).cache())[0].title,
+                queries=1
+            )
+        )
+        with self.assertNumQueries(0):
+            self.assertEqual('Changed', list(Category.objects.filter(pk=1).cache())[0].title)
+
+    def test_nested_nested_atomic_block_change_and_invalidate_all(self):
```
Suor: Just remove this test, `invalidate_all()` should not be used like this anyway.

tests/tests.py
```diff
@@ -938,3 +952,425 @@ def test_invalidate_model_with_geometry(self):
...
+                self.assertEqual(
+                    'Django',
+                    return_from_other_thread(
+                        lambda: list(Category.objects.filter(pk=1).cache())[0].title
+                    )
+                )
+                obj = list(Category.objects.filter(pk=1).cache())[0]
+                obj.title = 'Changed'
+                obj.save()
+                invalidate_model(obj.__class__)
```
Suor: If you really want to test `invalidate_model()` then you should use non-invalidating update:
```python
Category.objects.filter(pk=1).update(title='Change')
```
But I think this test is unnecessary.

Suor:
Overall your test set is weird - you should never need more then 2 levels of `atomic()`. Because 2 levels are sufficient to test both transactions and save points.

cacheops/transaction.py
jhillacre: Ok, I gotta go read more funcy docs, seems like it does some pretty neat stuff.

tests/tests.py
jhillacre: `TransactionTestCase` unwraps the tests from transaction.atomic(). If you run the tests with `TestCase` in, many will fail as they don't take in account the differences in queries and invalidtion delays that happen in transaction.

tests/tests.py
jhillacre: This is related to switching to `TransactionTestCase`. With `TestCase` every test wrapped in `atomic()`, the first line goes to SQL always as `BEGIN TRANSACTION` which django prepends with `SET SQL_AUTO_IS_NULL = 0'`.

jhillacre: Here is a link to a relevant django issue re: `SQL_AUTO_IS_NULL = 0`
https://code.djangoproject.com/ticket/24675

cacheops/transaction.py
jhillacre: Is there a better way to always get the generator version?
```python
from funcy.cross import PY3, PY2

if PY2:
    from funcy import iflatten as flatten, ikeep as keep
if PY3:
    from funcy import flatten, keep
```
...
```python
    def __iter__(self):
        return keep(flatten(self._queue, follow=is_list))
```

Suor: From [here](http://funcy.readthedocs.org/en/stable/python3.html#writing-cross-python-code):
```python
from funcy.py2 import ikeep, iflatten, is_list
```
However, you won't need this at all after `TransactionQueue` refactor.

jhillacre: Ok, pressing on.

tests/tests.py
jhillacre: removed in 25bb791e

tests/tests.py
jhillacre: removed in b3ce46e7

tests/tests.py
jhillacre: changed in b148d3f3

tests/tests.py
jhillacre: changed in b726baa1

tests/tests.py
jhillacre: removed in 25bb791e

tests/tests.py
jhillacre: removed in commits 7aadeea5 and b726baa1

tests/tests.py
jhillacre: Supported query number counting for other threads, which didn't have access to the test. Removed in b726baa1

cacheops/transaction.py
jhillacre: These changes were made in 53795eb1

cacheops/transaction.py
jhillacre: > Simplify this class interface to .begin(), .commit() and .rollback()

Changed in 53795eb1

Switched from nested contexts to stacked contexts in d8ca4788

cacheops/transaction.py
jhillacre: Changed as part of 53795eb18

cacheops/query.py
jhillacre: Switched in 4f651061
cache_on_save code got in_transaction() test in 438d9b13

cacheops/query.py
jhillacre: changed in fad38aa2

cacheops/invalidation.py
jhillacre: Sorry about this. fixed in 903c4bf7

jhillacre:
> Overall your test set is weird - you should never need more then 2 levels of atomic(). Because 2 levels are sufficient to test both transactions and save points.

With your feedback, tests should now be cut down much more to just testing if invalidation queueing works. Hoping you find they read better as well.

tests/tests.py
jhillacre: Will do in the morning. I think this is the only outstanding item from your feedback so far.

cacheops/transaction.py
```diff
@@ -0,0 +1,88 @@
...
+            self.in_transaction = False
+
+    def append(self, item):
+        self._queue[-1].append(item)
+
+_transaction_queue = TransactionQueue()
+
+
+def in_transaction():
+    return _transaction_queue.in_transaction
```
Suor:
```python
return bool(_transaction_queue._queue)
```
No need to track it manually.

cacheops/transaction.py
```diff
@@ -0,0 +1,88 @@
...
+            self.in_transaction = False
+
+    def rollback(self):
+        self._queue.pop()
+        if not self._queue:
+            # transaction
+            self.in_transaction = False
+
+    def append(self, item):
+        self._queue[-1].append(item)
```
Suor: This is actually `TransactionState` internally using stack, not queue using queue.

tests/tests.py
```diff
@@ -938,3 +951,120 @@ def test_invalidate_model_with_geometry(self):
...
+        except Exception as e:
+            # make sure other thread exceptions make it out to the test thread, instead of having
+            # None returned and the test thread continuing.
+            return e
+        finally:
+            # django does not drop postgres connections opened due to new threads. results in
+            #  Postgres complaining about connected users when django tries to delete test db
+            #  See https://code.djangoproject.com/ticket/22420#comment:18
+            from django.db import connection
+            connection.close()
```
Suor: `wrapper()` function is never used.

tests/tests.py
```diff
@@ -938,3 +951,120 @@ def test_invalidate_model_with_geometry(self):
...
+class IntentionalRollback(Exception):
+    pass
+
+
+class TransactionalInvalidationTests(BaseTestCase):
+    fixtures = ['basic']
+
+    def test_atomic_block_change(self):
+        with atomic():
+            self.assertTrue(in_transaction())
```
Suor: `in_transaction()` is internal interface and should not be tested.

tests/tests.py
jhillacre: using wrapper now in 6b214457. I also removed the finally part since it clearly wasn't needed (at least, anymore)

cacheops/transaction.py
jhillacre: renamed in 8f739650

tests/tests.py
jhillacre: removed in caa21bb2

cacheops/transaction.py
jhillacre: changed in 8cc1353c

tests/tests.py
jhillacre: moved in e766a44

tests/tests.py
jhillacre: The error encountered in this Travis CI build is what this `finally:` is there to deal.  https://travis-ci.org/Suor/django-cacheops/jobs/88385111
re-added in 8c62d5a

Suor:
Merged, thanks for this, it will be an important feature.

jhillacre:
Thanks for all the time you spent with this. Glad to see it merged.

Issue #173, opened 2015-11-17
No invalidation for intermediate models in long join chain


prokaktus:
Problem description:
Cacheops caches user's permissions and ignores Group permissions changes. For example, If I add permission to Group, `user.get_all_permissions()` (where `user` is some instance of `get_user_model()`) still return old queryset. I should manually invalidate Permission model to get correct response. I've tried to breakpoint issue, but looks like all signals works correctly (called pre- and post-save on Group and m2m_changed on GroupPermission).

Also I've wrote the test on this case.

Suor:
This issue in the end has nothing to do with m2m. Look at [this simplified test](https://github.com/Suor/django-cacheops/blob/c9a847e9d720c66b31226b4adb7872f3b30fbe44/tests/tests.py#L591). 

The queryset there has post -> category link in it, which could be broken:
```python
Extra.objects.filter(post__category__title=title)

# Break the link
extra.post.category_id = 2
extra.post.save()
```

The issue is fundamental - we can't know `category_id` or `post_id` from queryset conditions and thus can only invalidate on all Post changes or none. Both approaches have disadvantages. First invalidates too much, second has this issue.

Cacheops traditionally chooses not to invalidate on Post changes here. Actually pre 2.0 cacheops didn't invalidate on any joined conditions. I am not sure what cacheops should do here, should I leave it as is or add superfluos invalidation.

prokaktus:
Thank, @Suor!

Second approach have very serious problem: it can brings implicit errors into server logic. For example, you can consider my test example: changing group permissions in fact doesn't affect real user permissions and authorization. And this is just one example, there are could be more.

If you'll decide to choose second approach, it would be great to explicit indicate potential problems and suggest recommended way to bypass them. Maybe it should be signal handler, which will call invalidation function for an entire related model.

Suor:
I decided to invalidate on all intermediate tables. This is question of implementation now. Should definitely go in before 3.0.

PR #174, opened 2015-11-17
m2m for Group and Permission

prokaktus:
Test for issue #173

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -104,6 +104,7 @@ CACHEOPS = {
     'tests.*': {},
     'tests.noncachedvideoproxy': None,
     'tests.noncachedmedia': None,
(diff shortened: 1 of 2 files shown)


Suor:
`user.get_all_permissions()` uses local cache in `_perm_cache` attribute and you are reusing same user object. Also, number of queries can't demonstrate logic failure and this is what you are talking about here. Lastly you are not even using cache in your test, so everything behaves exactly as without cacheops.

prokaktus:
@Suor thanks for your reply! I've put some modification into test, so now it should show my intentions more clearly. My concern is that change auth.group doesn't invalidate auth.permission and I still get old permissions. Manual call for `./manage.py invalidate auth.permission` solving problem, but preferably auth.permission should be invalidated automatically.

PR #179, opened 2016-02-14
multi-table inheritance invalidation

KevinGrahamFoster:
I'm just getting to know this package, and love it, but I definitely need multi-table inheritance invalidation... Here's my rough stab at it. It's a crazy busy weekend of work for me so this pull request should be cleaned up, and tests added, but I was wondering if you could take a look and let me know if something like this is suitable, or if there are issues with it. If I am wrong on anything, let me know and I will clean it up.


Suor:
You should also consider going from the other end: when building DNF replicate each conjuction for all the tables in an inheritance tree, appropriately chunking of extra fields. E.g.:
```python
class Place(models.Model):
    name = models.CharField(max_length=50)
    address = models.CharField(max_length=80)

class Restaurant(Place):
    serves_hot_dogs = models.BooleanField(default=False)
    serves_pizza = models.BooleanField(default=False)

class IndianRestaurant(Restraunt):
    serves_curry = models.BooleanField(default=True)

qs = Restraurant.objects.filter(name='Villa', serves_pizza=True)
```
Now `dnf(qs)` will be:
```python
[('app_restaurant', [[('name', 'Villa'), ('serves_pizza', True)]])]
```
For mutiple-inheritance to work this should be:
```python
[
    ('app_place', [[('name', 'Villa')]]),
    ('app_restaurant', [[('name', 'Villa'), ('serves_pizza', True)]]),
    ('app_indianrestaurant', [[('name', 'Villa'), ('serves_pizza', True)]]),
]
```
Note how `serves_pizza=True` is cut for `Place`. This way invalidation won't change at all, no need to fetch all the related objects and no need to make lots of redis roundtrips. This won't add extra redis requests on cache write either, as it already expects DNF with multiple tables anyway.

The downside is that we loose some invalidation granularity, in example above any place named `'Villa'` change will lead to invalidation even if it's not even a restaurant.

KevinGrahamFoster:
Awesome, thanks for the comments. I'll get to them sometime this week, and bug you if I have questions (once again, just getting to know the package :+1: ).

Suor:
Another thing is that there are 2 tests already availabe for this. They are marked as expected failures now.

Suor:
Any progress on this?

jdotjdot:
We need this functionality as well, and I'm considering having someone on our team add this.  Have there been any major enough changes to cacheops that at this point a different approach than this should be taken?

Suor:
There are still two approaches here:
- Kevins original one,
- the one I described above.

I consider my approach superior), being simpler and more performant. Having a bit less invalidation granularity, but no change to invalidation procedure at all.

You can validate both approaches in the context of your own app/use case.

KevinGrahamFoster:
Yes, apologies for the silence, I completely moved away from this as it wasn't required for my project anymore.  I'd be willing to help finish this however if I can be of assistance.

mrmachine:
Not sure if it's useful, but here was my attempt at adding MTI support to an older version of cacheops that we're using in a Django 1.6 project:

https://github.com/ixc/django-cacheops/commit/e19968eb2d3ffa49e4133f6de60bb47f02067179

babus:
@Suor any updates please?

Suor:
Nobody is interested enough to finish this.

PR #182, opened 2016-02-26
Note about security

erthalion:
I'm not sure, but probably it will be nice to add note about potential security problems, because `pickle` can load some malicious data from redis in case if it was compromised due of  the misconfiguration. Or is `django-cacheops` not affected by this for some reason?


Suor:
If you provide write access to your redis to whoever wants it then you are already screwed.

PR #183, opened 2016-03-22
Allow redis connection settings to be specified by a URL

timsavage:
Simple change that allows for the redis connection setttings to be optionally specified using a redis:// url rather than a full dictionary, the url is parsed using `Redis.from_url`.

This is consistent with django-redis, celery.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,6 +56,9 @@ Setup redis connection and enable caching for desired models:
         'unix_socket_path': '' # replaces host and port
     }
 
+    # Alternatively the redis connection can be defined using a URL:
+    # CACHEOPS_REDIS = "redis://localhost:6379/1"
+
     CACHEOPS = {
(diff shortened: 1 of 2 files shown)


Suor:
Looks good, an example in README would be even better :)

timsavage:
Added note into readme.

There is one caveat in that the socket_timeout cannot be configured via a URL (this is a limitation of py-redis). Will submit a patch to the py-redis to resolve this. No code changes would be required on the part of django-cacheops to support additional parameters other than adding more details into README.

timsavage:
Pull request to handle `socket_timeout` in the URL was accepted into redis-py.

maryokhin:
This was like the only thing I was missing in this library, awesome :tada:

maryokhin:
You should also update the README example probably, for new users.

Issue #184, opened 2016-03-24
Cacheops and django-multidb-router


dpetzold:
Hello Suor!

We are running this error:

```
ValueError: Can't do subqueries with queries on different DBs.
  File "django/core/handlers/base.py", line 112, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 499, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 69, in view
    return self.dispatch(request, *args, **kwargs)
  File "django/utils/decorators.py", line 29, in _wrapper
    return bound_func(*args, **kwargs)
  File "core/decorator.py", line 105, in decorator
    return func(request, *args, **kwargs)
  File "django/utils/decorators.py", line 25, in bound_func
    return func(self, *args2, **kwargs2)
  File "ccauth/view_mixins.py", line 10, in dispatch
    return super(AgencyAdminRequiredMixin, self).dispatch(request, *args, **kwargs)
  File "newrelic/hooks/framework_django.py", line 872, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 87, in dispatch
    return handler(request, *args, **kwargs)
  File "django/views/generic/detail.py", line 111, in get
    context = self.get_context_data(object=self.object)
  File "dashboard/views.py", line 176, in get_context_data
    alerts = self.build_alerts(context_data)
  File "dashboard/views.py", line 194, in build_alerts
    alerts.extend(self.get_task_comments())
  File "dashboard/views.py", line 352, in get_task_comments
    @cached_as(comments_qs)
  File "cacheops/query.py", line 82, in cached_as
    key_extra = [qs._cache_key() for qs in querysets]
  File "cacheops/query.py", line 149, in _cache_key
    sql, params = self.query.get_compiler(self._db or DEFAULT_DB_ALIAS).as_sql()
  File "django/db/models/sql/compiler.py", line 94, in as_sql
    where, w_params = self.query.where.as_sql(qn=qn, connection=self.connection)
  File "django/db/models/sql/where.py", line 103, in as_sql
    sql, params = child.as_sql(qn=qn, connection=connection)
  File "django/db/models/sql/where.py", line 401, in as_sql
    raise ValueError("Can't do subqueries with queries on different DBs.")
```

the problem occurs under the following circumstances with multidb:

```
# routes to a slave
categories = Category.objects.filter(title='Python')
[...]

PR #185, opened 2016-03-24
Fix for #184.

dpetzold:


Suor:
Fixed with no tests in d78eb9771a89a08c0b02361c24b609d25e8faa91, don't want new dependency.

PR #186, opened 2016-03-24
Support for sending the age with the cache_read signal.

dpetzold:
Hi Suor,

This updates the cache_read signal to send the age if the setting CACHEOPS_SEND_AGE is set to True. I made it a setting since adds an additional redis call (ttl) with the get but it should be okay for everyone as the request is pipelined. Using the age we have been able to generate graphs like this using New Relic Insights.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -568,7 +568,9 @@ Keeping stats
 -------------
 
 Cacheops provides ``cache_read`` signal for you to keep stats. Signal is emitted immediately after each cache lookup. Passed arguments are: ``sender`` - model class if queryset cache is fetched,
-``func`` - decorated function and ``hit`` - fetch success as boolean value.
+``func`` - decorated function
+``hit`` - fetch success as boolean value.
+``age`` - The age of the object if CACHEOPS_SEND_AGE is True.
 
 Here is simple stats implementation:
 
--- a/cacheops/conf.py
+++ b/cacheops/conf.py
@@ -15,6 +15,7 @@ class Settings(object):
     CACHEOPS = {}
     CACHEOPS_LRU = False
     CACHEOPS_DEGRADE_ON_FAILURE = False
+    CACHEOPS_SEND_AGE = False  # Implies CACHEOPS_DEGRADE_ON_FAILURE
     FILE_CACHE_DIR = '/tmp/cacheops_file_cache'
     FILE_CACHE_TIMEOUT = 60*60*24*30
 
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -51,6 +51,27 @@ def cache_thing(cache_key, data, cond_dnfs, timeout):
     )
 
 
+def get_cache_data(cache_key, timeout, sender=None, func=None):
+
+    if settings.CACHEOPS_SEND_AGE:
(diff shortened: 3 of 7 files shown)


Suor:
Looks too specific to merge. Will leave this here, to see if more people like it.

brianz:
This seems quite useful to me. Support for detailed monitoring isn't specific at all IMO.

gposton:
I'd really like to see this get in as well.

Suor:
As far as I understand, you guys have single use case, so it's specific to your project so far. Code complication will, however, make project development harder for everyone for all foreseeable future. So I remain unconvinced.

There are components to that though. First, I don't see how this is useful, so graphs are pretty and we all love numbers, but how do you act basing on it? Second, this probably could be implemented outside of cacheops by somehow reading redis, do you really need this realtime?

gposton:
Great questions @Suor.  We set cache times based on our models.  I think this is pretty standard in cachops.  From the data we gather from the cache age, we can determine at what times after a cache key is created that the number of hits on that key decrease to a point where the storage of the key is no longer valuable.  This allows us to fine tune our TTLs by model.  In our high traffic, high caching environment, this enables us to tune/optimize the TTLs by model, which reduces our memory footprint in our cache cluster.  This in turn enables us to reduce the instance size of our cluster in AWS and significantly reduce associated infrastructure costs.

Suor:
So you don't need this anymore?

dpetzold:
Hi,

We could still use this. I was just cleaning up our repo as that change has been merged our master branch.

Thank you,

gposton:
@Suor, any additional thoughts based on my comments... any options to clean up the code or simplify it to get this in?

Suor:
I agree that this is generally usefull. There are issues with current approach though:

1. Extra setting.
2. `CACHEOPS_SEND_AGE` implies `CACHEOPS_DEGRADE_ON_FAILURE` which makes no sense.
3. Calculating age as `timeout - ttl` is not always correct because `timeout` could be different for same request. 
4. Overheads on both redis and django sides.

I suggest instead of using `TTL` command just pack timestamp when writing to cache:
```python
cache_data = pickle.dumps((data, time.time()), -1)
```
This will have zero overhead on redis side and small one on python side, so setting won't be needed. This will solve 1,3 and 4 (the last one partially).

To solve 2 just don't add methods to `SafeRedis`. If you still need extra methods on redis instance then create a subclass with conditionally overriden `.get()`:
```python
class CacheopsRedis(redis.StrictRedis):
    if settings.CACHEOPS_DEGRADE_ON_FAILURE:
        get = handle_connection_failure(redis.StrictRedis.get)

    def some_method(self):
        ...
```
And use it instead of `StrictRedis`.

gposton:
Good suggestions...  Thanks @Suor!  We'll get working on making the changes.

nicwolff:
If we could pass the Redis client base class in as a setting that would let us maintain customizations like this more easily without forking – I've just opened a PR for that here https://github.com/Suor/django-cacheops/pull/282.

Suor:
@nicwolff no this can't be solved with custom RedisClient, at least right now, since signal is sent outside of redis client.

PR #189, opened 2016-04-12
Added code to have the setting CACHE_OPS_KEY - bala

subramanyamVemu:
This is to have a user defined key string or a function that will be added before the key <  userdefined_key >as:xxx or < userdefined_key  >p:xxx 

user has to define a string or a function with CACHE_OPS_KEY

Please let me know your comments


subramanyamVemu:
A user can override the default key by having the following in settings

def get_key(default_key=None):
    if default_key is None:
        return None
    import datetime
    return datetime.datetime.now().strftime("%Y%m%d%H") + default_key

# assign a function
CACHE_OPS_KEY = get_key

subramanyamVemu:
I am not sure about the key -  conj:xx whether we want to delete the object from memory

Suor:
Not sure what you are trying to do here.

Suor:
Anyway, using separate redis instances is recommended. Using separate redis dbs could be used as compromise.

BTW, your code most probably won't work. There are not only cache keys, but conj keys and schemes. Also keys are also accessed through lua scripts.

Suor:
Closing this since this approach is not going to work and the lack of interest.

PR #195, opened 2016-05-24
Invalidation Django signals

ktkiiski:
Hi Suor,

This branch adds four (4) new Django signals for cache invalidation actions, similar to the existing `cache_read` signal:
- `invalidation_all`: the signal is sent from `invalidate_all` call
- `invalidation_model`: the signal is sent from `invalidate_model` call
- `invalidation_obj`: the signal is sent from `invalidate_obj` call
- `invalidation_dict`: the signal is sent from `invalidate_dict` call (undocumented, because the function itself is undocumented.)

We are ourself using these signals for collecting statistics about the number of invalidations, for different models.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -585,8 +585,14 @@ Tags work the same way as corresponding decorators.
 Keeping stats
 -------------
 
-Cacheops provides ``cache_read`` signal for you to keep stats. Signal is emitted immediately after each cache lookup. Passed arguments are: ``sender`` - model class if queryset cache is fetched,
-``func`` - decorated function and ``hit`` - fetch success as boolean value.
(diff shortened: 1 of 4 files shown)


Suor:
Why do you need invalidation call stats?

ktkiiski:
@Suor: To see if some models are invalidated significantly more than other models, or if the invalidation happens very often. This helps optimization, allowing us to better choose where to use caching or if there are actions that cause too much invalidation.

Suor:
Will it be enough to have single `cache_invalidate` signal with `sender` and `obj_dict` args?
```python
invalidate_dict(model, obj_dict) -> cache_invalidate.send(sender=model, obj_dict=obj_dict)
invalidate_obj() calls invalidate_dict()
invalidate_model(model) -> cache_invalidate.send(sender=model, obj_dict=None)
invalidate_all() -> cache_invalidate.send(sender=None, obj_dict=None)
```

Suor:
Add in fc9b850f5fac691796c7f94530dd6ab5b93ffaf8.

PR #202, opened 2016-08-03
Add test for caching annotations.

EvgeneOskin:


Suor:
Several issues with this.

1. Tests fail.
2. `lambda: Post.objects.all()` is not a change, no wonder this test fails.
3. Caching not tested only invalidation.
4. Tests are superflous, just test caching and test invalidaiton.

EvgeneOskin:
Thank @Suor for your review and for your remarks.

If fix tests, I've changed it to test caching and invalidation.

At the moment test fails in this one:
```python 
Category.objects.annotate(posts_count=Count('posts'))  # First hit
Post.objects.create(title='Django', category=Category.objects.all()[0])  # This call should invlidate cache.
Category.objects.annotate(posts_count=Count('posts'))  # It should be the second hit. But it use case
``` 
So, I expect that creating new `Post` should invalidate the annotated query cache.

Suor:
Merged, removed all tests but one. BTW, only single test was failing because you forget to turn on caching.

EvgeneOskin:
Awesome👍 thank you @Suor!

PR #209, opened 2016-09-02
Support for django ATOMIC_REQUESTS

ihucos:
When using Django's ATOMIC_REQUESTS django-cacheops silently has no effect since caching is disabled inside transactions (see https://github.com/Suor/django-cacheops/pull/171). 

Caching inside transactions is tricky, since we don't know if the view we have from the database will be commited at all, so it simply gets skipped. This pull request takes that strategy a little further, only skipping caching if a current transaction has changes at all. In a typical web app most requests are read only so with this PR they can profit from django-cacheops when Django's ATOMIC_REQUESTS is used.

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -97,8 +97,6 @@ CACHEOPS = {
     'tests.local': {'local_get': True},
     'tests.cacheonsavemodel': {'cache_on_save': True},
     'tests.dbbinded': {'db_agnostic': False},
-    'tests.genericcontainer': {'ops': ('fetch', 'get', 'count')},
-    'tests.All': {'ops': 'all'},
     'tests.*': {},
     'tests.noncachedvideoproxy': None,
     'tests.noncachedmedia': None,
--- a/tests/tests_transactions.py
+++ b/tests/tests_transactions.py
@@ -111,14 +111,25 @@ class TransactionSupportTests(TransactionTestCase):
         self.assertEqual('Django', get_category().title)
         self.assertEqual('Django', run_in_thread(get_category).title)
 
-    def test_dirtiness_with_nested_transaction(self):
-        obj = get_category()
-        self.assertFalse(transaction_state.is_dirty())
+    def test_use_cache_inside_transactions(self):
         with atomic():
-            self.assertFalse(transaction_state.is_dirty())
+            with self.assertNumQueries(1):
+                get_category()
+            with self.assertNumQueries(0):
+                get_category()
+            with atomic():
+                with self.assertNumQueries(0):
+                    get_category()
+
+            obj = get_category()
             obj.title += ' changed'
             obj.save()
-            self.assertTrue(transaction_state.is_dirty())
-            with atomic():
-                self.assertTrue(transaction_state.is_dirty())
-        self.assertTrue(transaction_state.is_dirty())
+
+            with self.assertNumQueries(1):
+                get_category()
+            with self.assertNumQueries(1):
+                get_category()
+        with self.assertNumQueries(1):
+            get_category()
+        with self.assertNumQueries(0):
+            get_category()
--- a/tests/tests_transactions.py
+++ b/tests/tests_transactions.py
@@ -121,3 +121,4 @@ class TransactionSupportTests(TransactionTestCase):
             self.assertTrue(transaction_state.is_dirty())
             with atomic():
                 self.assertTrue(transaction_state.is_dirty())
+        self.assertTrue(transaction_state.is_dirty())
--- a/tests/tests_transactions.py
+++ b/tests/tests_transactions.py
@@ -6,7 +6,6 @@ from django.db.transaction import atomic
 from django.test import TransactionTestCase
 
 from .models import Category
-from cacheops.transaction import transaction_state
 
 
 class ThreadWithReturnValue(Thread):
@@ -56,12 +55,9 @@ class TransactionSupportTests(TransactionTestCase):
 
     def test_atomic(self):
         with atomic():
-            self.assertFalse(transaction_state.is_dirty())
             obj = get_category()
-            self.assertFalse(transaction_state.is_dirty())
             obj.title = 'Changed'
             obj.save()
-            self.assertTrue(transaction_state.is_dirty())
             self.assertEqual('Changed', get_category().title)
             self.assertEqual('Django', run_in_thread(get_category).title)
         self.assertEqual('Changed', run_in_thread(get_category).title)
--- a/tests/tests_transactions.py
+++ b/tests/tests_transactions.py
@@ -110,3 +110,14 @@ class TransactionSupportTests(TransactionTestCase):
             self.assertEqual('Django', run_in_thread(get_category).title)
         self.assertEqual('Django', get_category().title)
         self.assertEqual('Django', run_in_thread(get_category).title)
+
+    def test_dirtiness_with_nested_transaction(self):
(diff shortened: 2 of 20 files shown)


ihucos:
Any timeline, when this will possibly get merged?

Suor:
Still not sure this is a good idea or whether it's worth the change.

ihucos:
What are your concerns?

Suor:
ATOMIC_REQUESTS is a bad idea, what's the point supporting them?

ihucos:
Its not only about ATOMIC_REQUESTS, its about taking the current strategy of how to handle transactions in general one step further. There may be some other use cases, where you want to use caching inside transactions as long as possible. This PR also shows a little bit a direction for even better transaction support: We could make the transaction dirty thing more fine-grained. So that for example only specific models get dirty/invalidated and possibly use existing logic for that.

I think that the more django-cacheops caches the more useful it becomes, and the code in this PR plays nicely with existing django-cacheops code and concepts.

Also ATOMIC_REQUESTS is a core part of Django, that yes, has a specific advantage/drawbacks profile but makes sense in some use cases.

ihucos:
@Suor can you close or merge, so at the company I am working for we can plan better our caching approach?

Suor:
Hello, @ihucos. Sorry for slow pace of handling this issue. I needed to decide whether this should go in. 

Anyway, I now agree that this is useful. So we should clean up implementation and merge it. Here are my considerations:

1. No need to handle dirtiness separately, it could be derived from the state of the stack.
2. `uncommited_changes()` is too vague, use `.dirty` property on `transaction_state`. Import `transaction_state` object everywhere instead of it.
3. Decorating `invalidate_obj()` is duplication of effort.

ihucos:
@Suor nice, glad to hear its going to be merged!  Will do the changes.

Suor:
Any proress on this?

ihucos:
@Suor I am Sorry, there came some stuff in between, it is the next thing I am working on.

ihucos:
As I understand the point 1. it means we don't need to save the dirtiness in the `TransactionState` but can calculate it from its existing `_stack` attribute. That would mean checking if it is empty, so if any callbacks are supplied. But wouldn't that be a bad abstraction, since a callback could theoretical be something that is not related to invalidation. Inspecting the callbacks in `_stack` also seems not perfect to me.

Suor:
No need to worry beforehand and no need to introspect callbacks, if something is on top of the stack then transaction is dirty, end of story.

ihucos:
OK, I did it like this but have the feeling that it does not work with nested transactions.
E.g. Inside a transaction data is changed - the top of the stack will have callbacks so the transaction is dirty. Then a new transaction is opened before the last one is committed, the top of the stack will be empty and the transaction_state will not be dirty, but data coming from the database gets inherited from the dirty transaction.

Suor:
When savepoint is commited its callbacks are transfered to the one enclosing it (a new stack top), so transaction remains dirty. 

The only issue I see is that non-invalidating but db writing event like mass update or `cursor.execute()` will leave transaction clean. This is the same level of correctness we provide for caching of cause, but lesser level than we had for transactions before this change.

Suor:
One more thing about tests:

1. Inner workings like `transaction_state.is_dirty()` should not be used.
2. There should be separate test for a feature. You are now polluting correctness test.

The right way, to my thinking, is creating separate test to check for behavior you are trying to achieve: caching works before first write in transaction but not after that.

ihucos:
> When a savepoint is commited its callbacks are transfered to the one enclosing it (a new stack top), so transaction remains dirty. Callbacks is the dirtyness, same semantics is no coincidence.
Or in a nested transaction it doesn't get commited yet and the callbacks are behind on the stack. I made a failing unit test with what I mean.

> Inner workings like transaction_state.is_dirty() should not be used.
OK, what should I use instead?

> There should be separate test for a feature. You are now polluting correctness test.
ACK

Suor:
You are right, transaction could become dirty before entering savepoint. So
dirtyness logic should be "if anything on the stack then it's dirty".

Suor:
> OK, what should I use instead?

The right way, to my thinking, is creating separate test to check for behavior you are trying to achieve: caching works before first write in transaction but not after that. 

Hint: use `self.assertNumQueries()` to check whether cache is hit.

ihucos:
Yeah makes sense, I'll write the tests like that.

Defining dirty as the existence of callbacks is still not doing it, see the adapted failing unit test. Maybe consider the original `TransactionState` in this PR?

Suor:
You can't test transactions with `BaseTestCase` only with `TransactionTestCase`. This is probably why that fails.

tests/tests_transactions.py
```diff
@@ -106,3 +110,15 @@ def test_nested_rollback(self):
...
+        obj = get_category()
+        self.assertFalse(transaction_state.is_dirty())
+        with atomic():
+            self.assertFalse(transaction_state.is_dirty())
+            obj.title += ' changed'
+            obj.save()
+            self.assertTrue(transaction_state.is_dirty())
+            with atomic():
+                self.assertTrue(transaction_state.is_dirty())
+        self.assertTrue(transaction_state.is_dirty())
```
Suor: We are not in a transaction here.

Suor:
Looked at it the error is because the tests is wrong: we are not in a dirty state when we are out of transaction.

A transaction test should still be removed from `tests.py`.

ihucos:
Ahh, yeah of course.

ihucos:
I think its good to go, check it out

ihucos:
Last ci run failed because of random Input/output error

Suor:
There is still on issue with this that I mentioned before:

>The only issue I see is that non-invalidating but db writing event like mass update or cursor.execute() will leave transaction clean. This is the same level of correctness we provide for caching of cause, but lesser level than we had for transactions before this change.

It could be fixed by more patching or a setting. The setting would work for now, with default to old behavior.

tests/settings.py
```diff
@@ -75,7 +75,7 @@
             # Make in memory sqlite test db to work with threads
             # See https://code.djangoproject.com/ticket/12118
             'TEST': {
-                'NAME': '/dev/shm/cacheops_sqlite.db'
+                'NAME': '/tmp/cacheops_sqlite.db'
```
Suor: Why is this?

tests/settings.py
```diff
@@ -75,7 +75,7 @@
             # Make in memory sqlite test db to work with threads
             # See https://code.djangoproject.com/ticket/12118
             'TEST': {
-                'NAME': '/tmp/cacheops_sqlite.db'
+                'NAME': '/dev/shm/cacheops_sqlite.db'
```
Suor: Please rebase to erase this from history

Suor:
Please rebase it to erase dev changes.

ihucos:
Have to rebase on monday

ihucos:
Done.

Don't know why github now also shows commits that are not mine in this pull request.

Suor:
You probably forgot to pull origin.

Suor:
Merged, thanks.

Suor:
Renamed `CACHEOPS_TRANSACTION_SUPPORT` setting to `CACHEOPS_SMART_TRANSACTIONS`.

ihucos:
Great!

Suor:
Removed the setting, all transactions will be smart based on cursor patch.

Issue #217, opened 2016-11-14
django>=1.10.x


GuillaumeCisco:
Hello,

Just tested last version of django-cacheops and Django 1.10.3.
Steps : 
./manage.py invalidate all
reload server
./manage.py invalidate all

Still it does not work properly, list is not invalidated.

If I run again command `./manage.py invalidate all` list is obviously invalidated but if I want to remove an item, still, list is not invalidated.
Anyone in this case, did I do something wrong?

My settings file is:
```python
CACHEOPS_REDIS = {
    'host':  environ.get('REDIS_HOST', '127.0.0.1'),
    'port': environ.get('REDIS_PORT', '6379'),
    'db': 1,
    'socket_timeout': 4,
}

CACHEOPS_DEFAULTS = {
    'timeout': 60*10  # 10 minutes
}

CACHEOPS = {
    'auth.user': {'ops': 'get'},
    '*.*': {'ops': 'all'}  # caching for all models
}

```

Suor:
You need to be more specific. Whats is the list? Why it should be invalidated? What do you mean by removing an item?

GuillaumeCisco:
Thanks @Suor for your quick response.
I'll try to be more specific.

Just using django-cacheops with django rest framework, so it is an API.
My list refer to a list of items when make a GET requests.
Deleting an item of this list consist of making a DELETE request on an item.
Sorry for misunderstanding.

So I've juste made my test on the Web Browsable API of a django rest framework project.
When I create or delete objects, and after request my list, I got a cached version. Nothing seems invalidated.

Thank you,

Suor:
Cacheops doesn't operate in terms of lists, deletes and so on. I don't know what happens within django rest framework so can't say why it doesn't work. People have used cacheops with django rest framework before with no issues though.

Maybe you have some cache on top of cacheops? You can try executing corresponding queryset in a console before and after change and look if that invalidates.

GuillaumeCisco:
Yes you're right, I think I don't express myself correctly.
When I talk about list, it is related to `MyModel.objects.all()`, it returns a list of items.
As far as I understand, this operation is invalidated when we create/delete an item of this MyModel, right?

I've used django-cacheops with django-rest-framework and had no issues for  the past 3 years :)
Just checked I have no cache on top of caheops and django version 1.9.11 works perfectly with it.
[...]

PR #218, opened 2016-11-14
Add weak=False to connect

GuillaumeCisco:
Fix https://github.com/Suor/django-cacheops/issues/217


Suor:
How about writing a falling test?

GuillaumeCisco:
Sorry I don't have time for that.

Suor:
This doesn't look as a right fix. Cacheops doesn't connect local functions via signals but manager methods. This means your code destroys manager (wtf???) and so removes signals, but when new manager is probably created with possibly new signal handlers attached. With this magic happening and `weak=False` you'll get a memory leak because manager won't be freed and more importantly more and more handlers will be attached to same signals running invalidation several time for each event. This will lead to high CPU consumption and over time will hang up everything.

GuillaumeCisco:
Sorry, but I don't have time to investigate more.
I let this to others.
Hope this discovery helps.

Suor:
Looked once more at the code, signals are garanteed to not be installed more than once for single class. So worst case scenario won't happen, but still something weird is happening in your code. It can bite you some day.

PR #220, opened 2016-11-16
Allow Redis-keys prefixing

raphael-riel:
Hi! 
For our system, we have multiple instances of the same codebase running on different DBs, but all sharing the same Redis server. 

* Spinning up one Redis per instance is not an option due to costs involved. 
* Attributing a unique DB-Number to every instance was considered, but excluded because of complexity involved with our random/ephemeral instances created/deleted over time.

Last resort is prefixing every key we set in Redis to avoid mixing data and dealing with conflicts; as Django's `CACHES` settings offers.

Looks like this topic was already covered by #189, but I would like to offer you my version of this feature. With this PR, Lua scripts are covered.

Final result in Redis:

```
127.0.0.1:6379[3]> FLUSHDB
OK
127.0.0.1:6379[3]> SCRIPT FLUSH
OK

[... run Django ...]

127.0.0.1:6379[3]> KEYS *
1) "cacheops_rriel:schemes:languages_language"
2) "cacheops_rriel:conj:languages_language:code=fr"
3) "cacheops_rriel:conj:languages_language:id=1"
4) "cacheops_rriel:conj:languages_language:code=en"
5) "cacheops_rriel:q:da08b5bf1acb38ff212e1ec56d754f41"
6) "cacheops_rriel:conj:languages_language:id=2"
7) "cacheops_rriel:q:9304694f24ee38a3f98a09ad41a2fe0b"
8) "cacheops_rriel:q:4a02481e255582de17b6f95763d68893"
```

~Bonus, I wrote a Docker/Makefile for OSX Folks who want to run the test suite.
The SQLite's `/dev/shm/` memory trick is not available on OSX. Instead of patching the tests, I preferred to allow test to run in a controlled environment.~


Suor:
Not interested in docker files, this is to big of a dependency to run tests. Especially when the issue can be resolved with a simple if. And anyway this should not be mixed into unrelated thing.

As for the rest, approach looks too narrow. The obvious extension use-case: single installation and splitting cache on e.g. subdomain. Also, cacheops should not be littered with prefix code, changes should concentrate on python-redis borders.

raphael-riel:
@Suor The Dockerfile is an alternate way to run tests; in no way they replace the original (untouched) test setup. Will remove it.

As for the prefix, my first attempt was to allow a custom Redis connector in place of StrictRedis (subclassing it) so prefix was added the moment before gets/sets were sent to Redis. Unfortunately, this method wasn't working with the cache/invalidation Lua scripts. 

What about a custom Redis connector and keeping the `%PREFIX%` substitution in `load_script` and the Lua scripts?

Suor:
A substitution in scripts obviously won't work if prefix is defined dynamically.

There are several places where prefix needs to be used:
- cache_thing()
- invalidate_dict()
- invalidate_model()
- redis_client.getting()
- QuerySetMixin.iterator()

The last one to be dropped on next major release. Changes should not appear in cache key calculation code.

raphael-riel:
Thanks for the hints! I'll attempt a Refactor to limited key prefixing to fewer methods.

raphael-riel:
Finally, we will not be using CacheOps.
Thanks for you're time assisting me on this PR.

PR #228, opened 2017-02-22
Fix late invalidation for djangos on commit hook

ihucos:
The `self._no_monkey.__exit__` call triggers possible on commit handlers [1] when inside an transaction. In most of this cases this will get unnoticed because the transaction is marked dirty and the cache is not used anyway. But we triggered a celery worker in an on commit handler and got a race condition where sometimes we got old data since it was fetched before the invalidation.

[1] https://docs.djangoproject.com/en/1.10/topics/db/transactions/#django.db.transaction.on_commit

### DIFF
--- a/cacheops/transaction.py
+++ b/cacheops/transaction.py
@@ -66,13 +66,13 @@ class AtomicMixIn(object):
         self._no_monkey.__enter__(self)
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self._no_monkey.__exit__(self, exc_type, exc_value, traceback)
         connection = get_connection(self.using)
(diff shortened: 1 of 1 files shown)


Suor:
This line can't be just moved like that - `connection.needs_rollback` value can change affecting commit/rollback logic.

Suor:
Looking a bit more into it I wonder if there are more issues to current code. Maybe moving from patching Atomic to patching connection is required.

Suor:
Another thing is that running invalidtion before commit is an obvious race. Someone might just cache old data inbetween and that will hang there until next change.

ihucos:
Yeah, True, I thought there was a very quick fix, hmmm...

PR #229, opened 2017-02-22
Use cacheops on commit handlers instead of djangos

ihucos:
How about that one?
Related: https://github.com/Suor/django-cacheops/pull/228


Suor:
This might work. There are however an issue of callbacks gotten mixed up when savepoints are involved. 

Also, are you sure semantics is the same?

ihucos:
Yeah, maybe that solves one issue but may cause other more subtle ones. I'll explore some other fixes. I am not sure sematics is the same.

PR #230, opened 2017-02-24
Run on_commit hooks after invalidation

ihucos:
Ok, I think that one should be good.
We are not changing on_commit behaviour. As I understand it, djangos on_commit hooks are called when a database transaction is commited, but nested Atomics actually leads to the usage of savepoints and not nested transactions. So the behaviour of cacheops on transaction success and djangos on_commit callbacks is different. The only assumption we do now, is that when djangos on_commit hook is called, cacheops transaction_state also needs to be commited.

I actually tried to maintain the `transaction_state` by only patching BaseDatabaseWrapper and that did not work. I think there is information that cacheops needs that can only be seen by the the higher level `Atomic` in a sane way.

It would be nice to also support django-transaction-hooks when used prior to django 1.9

Related https://github.com/Suor/django-cacheops/pull/228 https://github.com/Suor/django-cacheops/pull/229

EDIT:
Tell me if its worth fixing Django 1.7 support

### DIFF
--- a/cacheops/transaction.py
+++ b/cacheops/transaction.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
-import six
 import threading
 
-from funcy import wraps, once
+import six
+from django.db.backends.base.base import BaseDatabaseWrapper
 from django.db.backends.utils import CursorWrapper
-from django.db.transaction import get_connection, Atomic
+from django.db.transaction import Atomic, get_connection
+from funcy import once, wraps
 
 from .utils import monkey_mix
-
 
 __all__ = ('queue_when_in_transaction', 'install_cacheops_transaction_support',
            'transaction_state')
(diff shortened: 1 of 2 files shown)


Suor:
I don't like this implementation, it's now split across 2 mixins and some flag added, too much mess. And sure Django 1.7 should work.

ihucos:
I think that even with the code not being the most beautiful, the implementation is robust.
The bug affects a quite common scenario, starting a celery task in an on_commit hook (that is also mentioned in the Django docs [1] as an example usage). It should definitely be fixed. I looked into all approaches I could find and this is the best I could come up with after some days. If you know something better we can try it. We could also enable this patch with a flag to mitigate possible bugs or side effects. In any case, tell me how you want to move forward with this and If I should put more effort in this PR.

[1] https://docs.djangoproject.com/en/1.10/topics/db/transactions/#performing-actions-after-commit

ihucos:
@Suor at the company I am working we are forking with this PR for now, I'd be happy to work more on this PR or try another approach.

Suor:
I think there is an error here. If connection rolles back then flag is not set and `transaction_state` stack is not cleared.

Suor:
Should be fixed with 1ad6f95121f54abeef86552e263cac0e3f3a3329

ihucos:
> I think there is an error here. If connection rolles back then flag is not set and transaction_state stack is not cleared.
Ahhhhhh. Yes.

Thank you very much for looking into it and fixing !!!

PR #233, opened 2017-03-17
Fix clean_dnf sorting issue

### LINKED ISSUE #232: Issue when sorting conjunctions in DNF
Turns out that there is an issue when querying the same field multiple times when one of the query is for `__isnull=True`.

For example, given we have following models:

    class Site(models.Model):
        domain = models.CharField(max_length=127)

    class Article(models.Model):
        site = models.ForeignKey(Site, null=True)

The following query:

    Article.objects.cache().filter(Q(site__isnull=True) | Q(site__in=[site])).filter(site=site)

fails with `TypeError: unorderable types: int() < NoneType()`. The error happens while sorting conjunctions in `clean_dnf` function:

    map(sorted, cleaned)

where `cleaned` in this case is:

    [[('site_id', None), ('site_id', 1)]]

One way to fix this would be to cast values to strings before comparing them, like that:

    map(lambda l: sorted(l, key=lambda x: str(x)), cleaned)

I will submit PR with the above fix and relevant test case in a minute.

michalochman:
Fixes #232

### DIFF
--- a/cacheops/invalidation.py
+++ b/cacheops/invalidation.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import json
 import threading
+from builtins import str
 from funcy import memoize, post_processing, ContextDecorator
 from django.db.models.expressions import F
 # Since Django 1.8, `ExpressionNode` is `Expression`
@@ -99,4 +100,4 @@ def get_obj_dict(model, obj):
         elif isinstance(value, (F, Expression)):
             continue
         else:
-            yield field.attname, field.get_prep_value(value)
+            yield field.attname, str(field.get_prep_value(value))
--- a/cacheops/invalidation.py
+++ b/cacheops/invalidation.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import json
 import threading
-from builtins import str
 from funcy import memoize, post_processing, ContextDecorator
 from django.db.models.expressions import F
 # Since Django 1.8, `ExpressionNode` is `Expression`
@@ -100,4 +99,4 @@ def get_obj_dict(model, obj):
         elif isinstance(value, (F, Expression)):
             continue
         else:
-            yield field.attname, str(field.get_prep_value(value))
+            yield field.attname, field.get_prep_value(value)
--- a/cacheops/tree.py
+++ b/cacheops/tree.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-from builtins import str
 from itertools import product
 # Use Python 2 map here for now
 from funcy.py2 import map, cat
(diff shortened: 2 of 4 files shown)


Suor:
There are several issues with this patch:

1. Solution is inefficient in several ways. Main one is that values are converted to strings anyway in lua, so one can just convert them before sorting.
2. Problem is Python 3 only, so inefficiency for Python 2 users is even less acceptable.
3. Tests are too volumous. You don't need 2 new models, I don't think you even need one, just use one of the basic. Also, there is no need to create objects, this doesn't affect test anyway.

michalochman:
@Suor I will be happy to adjust the patch, but I have no idea what would be the best fix to this problem, any suggestions?

Suor:
Looked into it a bit more and it looks as not a practical issue see my comment in #232. This still could be fixed, though, to handle stupid queries generated by some complicated code. The easiest way is probably to use `str(value)` in `clean_conj()`.

michalochman:
Unfortunately, using `str(value)` in `clean_conj()` causes `test_invalidate_by_boolean` and `test_reverse_fk_same` to fail.

Suor:
This is because you need to prepare obj dict the same way for invalidation.

michalochman:
@Suor I have implemented it according to your guidance and removed unnecessary models from tests. Also, `str` is imported from `builtins` in order for the code to be Python 2-3 compatible. 

What do you think? See my comment in #232 for the real world example of the problem.

michalochman:
Seems that `builtins` is not available on Travis, let me investigate.

Suor:
The point of builtins that you don't need to import them :) Tests fail because in Python 2 builuins module is called `__builtin__`. Not that this matters anyway.

michalochman:
I now see that the project has `six` listed as it's dependencies so installing `future` seems pointless. Should I remove the import altogether and just use whatever `str` is available at runtime?

Suor:
Also, you will run into failing to `str()` unicode data - `.test_114()`.

I think we probably should try an alternative solution - filtering out dups in `clean_dnf()` or `clean_conj()`.

In both cases performance should be tested before accepting anything.

michalochman:
The reason I wanted `str` imported from `builtins` (coming from `future`) was meant to cover the unicode case, see http://python-future.org/compatible_idioms.html#unicode

---

Unless you can guarantee that the function removing duplicates always returns the same value for the following input:

    [[('attname', 'value'), ('attname', None)]]
    [[('attname', None), ('attname', 'value')]]

I am not sure that removing duplicates is a good idea, because they are not really "duplicates". What is unique here is a tuple of `(attname, value)` not just `attname`, don't you think? 

If that function was stable, it would return the first element of the inner list which is either `('attname', 'value')` or `('attname', None)` respectively. If it was not stable, it would return the inner items in arbitrary order.

Anyway, if this is not the issue, I found this benchmark: https://www.peterbe.com/plog/uniqifiers-benchmark and have tested it using `f5` function which was the fasted stable function under benchmark. The code and results are here: https://gist.github.com/michalochman/7d66b3fc92fee3cc519fb98d3d1a4b4c

Suor:
Conjs with duplicate fields and different values are useless, they will never cause invalidation. Dups with same values could be safely deduped with no resulting change in invalidation logic.

So removing dups should:
- completely remove any conj like `[['f1', None], ['f1', 'v'], ['f2', 'v2'], ...]`
- dedupe `[['f1', 'v'], ['f1', 'v'], ['f2', 'v2'], ...]` into `[['f1', 'v'], ['f2', 'v2'], ...]`

Suor:
Returning empty list is a bug. Empty list means empty conj = match all.

michalochman:
What do you mean by

> completely remove any conj like `[['f1', None], ['f1', 'v'], ['f2', 'v2'], ...]`

then?

Suor:
It should be removed from a list of conjs not replaced with empty
conjunction.

michalochman:
Oh, you're right – fixed.

Issue #234, opened 2017-03-19
is_sql_dirty matches on column name


icereval:
Transaction method `is_sql_dirty` matches on strings `update`, `delete` or `insert`. Causing false positives when column names match.

e.g.

`is_deleted`
`delete`
`deleted`

Suor:
Yes it is. This is ok, can cause superficial invalidation but won't hurt logic.

icereval:
In our case loading a User (w/ a column `is_deleted`) out of the database/cache on each authenticated request coupled with using Django's [Atomic Requests](https://docs.djangoproject.com/en/1.10/topics/db/transactions/#tying-transactions-to-http-requests) features causes cacheops to disable itself immediately.

Suor:
I will think about it

Suor:
So my considerations so far:

1. Atomic requests is a stupid idea, which I currently support though.
2. There are rational use cases which could be bitten by this.
3. Even though no logic issues stem from this, certain cases are still completely ruined.
4. Reading user from cache in your example won't really dirty the transaction as long as that sql is not executed. Won't probably help much anyway.

This means just dropping support for atomic requests won't help here :) So we come to the last consideration - regular expressions is slow. I don't like making everything slower, but I'll see what I can do.

PR #235, opened 2017-03-19
Regex search for dirty sql statements

icereval:
see #234


Suor:
Not interested. This simplification was intentional from the start. This is rare case, which doesn't even break logic just causes superfluous invalidation i.e. some inefficiency.

Issue #237, opened 2017-04-04
Cache is saving the object in presave without the model been in CACHEOPS configs


leonardoo:
i was looking with django debug toolbar, and a object that is not config in CACHEOPS was been searched and added, to the cachelist but this object was not in the cacheops list

Suor:
If `cache_on_save` is set then it will work, ops do not affect this. If that flag is not set then there should be no precaching. 

If you believe this is not how it works now then prove it with the test.

leonardoo:
i figure out why this happend, i use something like this

id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)

so this condition is always "instance.pk is not None" is always true in the _pre_save

Suor:
This is irrelevant've

PR #238, opened 2017-04-04
Don't presave models by pk if it's not in the cacheops config

leonardoo:
#237 with this pr, the object that is not in the cacheops config will not be search and presave

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -29,6 +29,7 @@ from .transaction import transaction_state
 from .signals import cache_read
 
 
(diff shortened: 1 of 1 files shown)


Suor:
`_pre_save()` is used to fetch old object state to properly invalidate on save. It is not used to cache anything. That is done in `_post_save()` and there is a check.

Suor:
So this is the real bug after all. One that has nothing to do with neither titles nor descriptions here and in #237.

PR #258, opened 2017-11-17
Sentinel

Tapo4ek:


Suor:
Hello and sorry for slow response. I don't mind adding some support of sentinel, would be good to hear your use case though.

Also, I have some questions here. Why are you using not documented class, wouldn't it better to use:
```python
sentinel = Sentinel([('localhost', 26379)], socket_timeout=0.1)
return sentinel.master_for(name, redis_class=CacheopsRedis)
```

Another thing is your use of configuration params, "specify both" looks wrong, when you need both anyway it's better to use only one:
```python
CACHEOPS_SENTINEL = {
    'locations': [('localhost', 23579)],
    'service_name': 'mycache',
}
```
This way we can later add any params we need, like `socket_timeout` above.

Do you have time to modify your PR?

Tapo4ek:
Hi!
Fixed by your comments.

PR #268, opened 2018-01-24
Issue 265

M1hacka:
Fixing Travis MySQL error in test


M1hacka:
Great, MySQL doesn't rollback DROP TABLE after transaction end... I'll fix it tomorrow

Suor:
Ah, DDL is not transaction safe there. Maybe the easiest fix is to skip
this test on MySQL.

24 янв. 2018 г. 21:42 пользователь "M1ha Shvn" <notifications@github.com>
написал:

Suor:
Fix for all these fixes. For linting too)

PR #269, opened 2018-01-26
Issue 265

M1hacka:
self._db in invalidated_update doesn't contain database in some cases (If router works, not strict using statement). I can't write test for it, but I catched it in my project.

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -387,7 +387,7 @@ class QuerySetMixin(object):
 
         # TODO: do not refetch objects but update with kwargs in simple cases?
         pks = {obj.pk for obj in objects}
-        for obj in chain(objects, self.model.objects.filter(pk__in=pks).using(self._db)):
+        for obj in chain(objects, self.model.objects.filter(pk__in=pks).using(clone.db)):
             invalidate_obj(obj)
         return rows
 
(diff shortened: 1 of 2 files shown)


Suor:
What some cases? Can you explain situation fuller?

26 янв. 2018 г. 19:52 пользователь "M1ha Shvn" <notifications@github.com>
написал:

M1hacka:
I use django-sharding. It determines database from filter() method params. When I call invalidated_update method, it saves data well (clone.db is correct database), but self._db is None. As a result, filter() after update fails with "table does not exist"

Suor:
If that determines for update why it fails for fetch?

26 янв. 2018 г. 21:25 пользователь "M1ha Shvn" <notifications@github.com>
написал:

Suor:
Ah, I see. Using `self.db` will fix it, but that might be wrong theoretically since new changed objects might have different db. In practice, though, `QuerySet.update()` wouldn't move objects to different database unless it has some custom implementation.

Using `clone.db` might also makes sense especially if we are consider slave lag - `clone` is for write queryset unlike `self` there. In both cases we defy router a chance route by pks in actual queryset.

Also we may just require specifying db in complex cases like this manually via `.using()`. Then current code will work.

Too many approaches.

Suor:
Cherrypicked it.

PR #280, opened 2018-04-10
Use Django bundled six to avoid a dependency

jdufresne:
Django ships with a bundled version of six. Can rely on this instead of including six as a separate dependency. Doing so results in one fewer dependency for package users that don't need six.

Django is committed to continued inclusion for backwards compatibility.


Suor:
I see little value here, what's the point having one dependency less? There could be issues though, since I won't be able to control six version and will have that depending on Django version.

PR #282, opened 2018-05-08
Make the base Redis client class a setting that defaults to StrictRedis

nicwolff:
We've been maintaining a fork of cacheops for some time which has fallen farther and farther behind your releases – if we could pass in the base Redis class we could keep most of our customizations in our Django application codebase, and it'd be much easier to keep cacheops up to date.

**This change adds a `CACHEOPS_CLIENT_CLASS` setting that can be used to pass a class name to import and use instead of `redis.StrictRedis`.** Generally I expect this will be a subclass of `StrictRedis` with some customizations or enhancements, but it could be any class with the same interface.

### DIFF
--- a/cacheops/conf.py
+++ b/cacheops/conf.py
@@ -18,6 +18,7 @@ class Defaults(namespace):
     CACHEOPS = {}
     CACHEOPS_PREFIX = lambda query: ''
     CACHEOPS_LRU = False
+    CACHEOPS_CLIENT_CLASS = None
     CACHEOPS_DEGRADE_ON_FAILURE = False
     CACHEOPS_SENTINEL = {}
 
--- a/cacheops/redis.py
+++ b/cacheops/redis.py
@@ -4,6 +4,7 @@ from contextlib import contextmanager
 import six
 
 from django.core.exceptions import ImproperlyConfigured
+from django.utils.module_loading import import_string
 
 from funcy import decorator, identity, memoize, LazyObject
 import redis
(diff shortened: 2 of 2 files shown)


Suor:
Have some questions here:

1. What issues are you going to solve with this?
2. Why parametrize subclass instead of resulting client class?

```python
@LazyObject
def redis_client():
    redis_class = settings.CACHEOPS_REDIS_CLASS or CacheopsRedis

    # ...
    if settings.CACHEOPS_SENTINEL:
        sentinel = ...
        return sentinel.master_for(
            # ...
            redis_class=redis_class,
            # ...
        )

    # Allow client connection settings to be specified by a URL.
    if isinstance(settings.CACHEOPS_REDIS, six.string_types):
        return redis_class.from_url(settings.CACHEOPS_REDIS)
    else:
        return redis_class(**settings.CACHEOPS_REDIS)
``` 

Here are [LazyObject docs](https://funcy.readthedocs.io/en/stable/objects.html#LazyObject).

nicwolff:
Thanks for your reply!

_What issues are you going to solve with this?_

We maintain a fork of cacheops to which we've added a configurable circuit breaker and weighted read-replica randomization. I'll be happy to open pull requests for them! But in the meantime it'd facilitate customizing cacheops while still keeping up with your changes.

_Why parametrize subclass instead of resulting client class?_

Ha, that's what everyone here asked! I didn't want to lose your `CacheopsRedis` improvements, and it seemed cleaner to pass objects one way rather than importing that class and sending a subclass back in as a setting. And maybe LazyObject scared me off. But I'll test and push that change ASAP.

nicwolff:
I've made those changes, but I'm still importing the class at runtime, so that the caller can subclass `CacheopsRedis` from here and pass the new class in as a setting without a circular import here.

Suor:
Ok merged, thanks for changes, you can open source your custom redis client(s) independently now. I wait for code too cool a bit before release, typically several days.

PR #284, opened 2018-05-26
added locking on get support for cached decorator as cached_as

### LINKED ISSUE #283: Enable cache locking for cached decorator like cached_as

chetandhembre:
fixes #283

### DIFF
--- a/cacheops/simple.py
+++ b/cacheops/simple.py
@@ -86,7 +86,7 @@ class RedisCache(BaseCache):
         self.conn = conn
 
     def get_with_lock(self, cache_key):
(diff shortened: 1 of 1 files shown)


chetandhembre:
All failed test cases are because of unable to create "create user cacheops" users in mysql.
I unable to understand how it is because of my code changes.
is there anything I can change in this pr to fix it?

Suor:
This won't work, you need to only release lock after you've written a new value to cache. MySQL issue could be caused by some changes to TravisCI.

Suor:
Another issue with lock for simple cache is that it won't work for anything but RedisCache, `self.conn` is not guaranteed to be there.

PR #293, opened 2018-07-24
Fix #292 : Give using parameter to invalidation functions

### LINKED ISSUE #292: Invalidation and multiple database
The invalidate functions in invalidation.py are designed to receive a db connection name (using parameter) but never receive it. Instead, DEFAULT_DB_ALIAS is always used. 
Exemple of such usages : 
- post_save, post_delete and m2m_changed receivers. 
- bulk_create and invalidated_updated

Svan70:
#292 
Give using argument to the invalidation functions call in query.py (invalidate_obj, invalidate_dict). 
I did not modify invalidate_* call in management. Not sure it is needed to do something in this file.

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -10,6 +10,7 @@ from .cross import pickle, md5
 import django
 from django.utils.encoding import smart_str, force_text
 from django.core.exceptions import ImproperlyConfigured
+from django.db import DEFAULT_DB_ALIAS
 from django.db.models import Manager, Model
 from django.db.models.query import QuerySet
 from django.db.models.sql.datastructures import EmptyResultSet
@@ -375,7 +376,7 @@ class QuerySetMixin(object):
         objs = self._no_monkey.bulk_create(self, objs, batch_size=batch_size)
         if family_has_profile(self.model):
             for obj in objs:
-                invalidate_obj(obj)
+                invalidate_obj(obj, using=self.db)
         return objs
 
     def invalidated_update(self, **kwargs):
@@ -391,7 +392,7 @@ class QuerySetMixin(object):
         pks = {obj.pk for obj in objects}
         new_objects = self.model.objects.filter(pk__in=pks).using(clone.db)
         for obj in chain(objects, new_objects):
-            invalidate_obj(obj)
+            invalidate_obj(obj, using=self.db)
         return rows
 
 
@@ -440,9 +441,10 @@ class ManagerMixin(object):
 
         # Invoke invalidations for both old and new versions of saved object
         old = _old_objs.__dict__.pop((sender, instance.pk), None)
+        using = kwargs.get('using', DEFAULT_DB_ALIAS)
         if old:
-            invalidate_obj(old)
-        invalidate_obj(instance)
+            invalidate_obj(old, using=using)
+        invalidate_obj(instance, using=using)
 
         # We run invalidations but skip caching if we are dirty
         if transaction_states[instance._state.db].is_dirty():
@@ -485,7 +487,8 @@ class ManagerMixin(object):
         """
         # NOTE: this will behave wrong if someone changed object fields
         #       before deletion (why anyone will do that?)
-        invalidate_obj(instance)
+        using = kwargs.get('using', DEFAULT_DB_ALIAS)
+        invalidate_obj(instance, using=using)
 
     def inplace(self):
         return self.get_queryset().inplace()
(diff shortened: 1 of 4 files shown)


Suor:
Thanks. Two things:
- why use `kwargs.get()` using should be always passed judging by Django docs
- can you test it?

Svan70:
I wasn't sure it was always passed so I used .get to avoid any problem ! I remove them. 
I wrote some tests. Are you ok with them ?

cacheops/query.py
```diff
@@ -524,18 +526,19 @@ def invalidate_m2m(sender=None, instance=None, model=None, action=None, pk_set=N
         instance_column, model_column = model_column, instance_column
 
     # TODO: optimize several invalidate_objs/dicts at once
+    using = kwargs['using']
```
Suor: I meant unpacking `using` in func signature.

tests/settings.py
```diff
@@ -96,7 +96,7 @@
 CACHEOPS = {
     'tests.local': {'local_get': True},
     'tests.cacheonsavemodel': {'cache_on_save': True},
-    'tests.dbbinded': {'db_agnostic': False},
+    'tests.dbbinded': {'ops': 'count', 'db_agnostic': False},
```
Suor: If you are calling `.cache()` then ops are ignored.

tests/tests.py
```diff
@@ -890,3 +893,98 @@ def test_invalidate_model_with_geometry(self):
         geom.save()
         # Raises ValueError if this doesn't work
         invalidate_obj(geom)
+
+
+class MultiDBInvalidationTests(BaseTestCase):
+
+    @override_settings(CACHEOPS_PREFIX=lambda q: q.db)
```
Suor: Should not be needed since this model is `'db_agnostic': False`

tests/tests.py
```diff
@@ -890,3 +893,98 @@ def test_invalidate_model_with_geometry(self):
...
+
+        db_binded = DbBinded()
+        DbBinded.objects.using('slave').bulk_create([db_binded])
+        with self.assertNumQueries(0):
+            DbBinded.objects.cache().count()
+        with self.assertNumQueries(1, using='slave'):
+            DbBinded.objects.cache().using('slave').count()
+
+    @mock.patch('cacheops.invalidation.invalidate_dict')
+    def test_bulk_update_call_invalidate(self, mock_invalidate_dict):
```
Suor: Unneeded test of the internals.

tests/tests.py
```diff
@@ -890,3 +893,98 @@ def test_invalidate_model_with_geometry(self):
...
+        Category.objects.using('slave').invalidated_update(title='update')
+        mock_invalidate_dict.assert_called_with(mock.ANY, mock.ANY, using='slave')
+
+    @override_settings(CACHEOPS_PREFIX=lambda q: q.db)
+    def test_post_save(self):
+        DbBinded.objects.cache().count()
+        DbBinded.objects.using('slave').cache().count()
+
+        obj = DbBinded()
+        obj.save(using='slave')
```
Suor: Testing `.save(using=...)` is ok, but testing implicit db should go first.

cacheops/query.py
```diff
@@ -440,9 +441,10 @@ def _post_save(self, sender, instance, **kwargs):
...
         old = _old_objs.__dict__.pop((sender, instance.pk), None)
+        using = kwargs['using']
         if old:
-            invalidate_obj(old)
-        invalidate_obj(instance)
+            invalidate_obj(old, using=using)
+        invalidate_obj(instance, using=using)
 
         # We run invalidations but skip caching if we are dirty
         if transaction_states[instance._state.db].is_dirty():
```
Suor: What happens if `using != instance._state.db`?

Suor:
Also look at failed checks in Travis, please.

cacheops/query.py
Svan70: Ok, it's better.

tests/tests.py
Svan70: That's why I used that class but I put that decorator because otherwise it doesn't work. If I correctly understand your code this is the reason : when you cache a queryset, the db_agnostic settings implies (as we wish) that cache key are different for the 2 query set because they use a different db connection. However, their prefix is the same because default CACHEOPS_PREFIX  returns an empty string. The problem is that when you invalidate the cache (in bulk_create for instance), the invalidation for each object is based only on the prefix, db_table and obj_value. Three are the same for 2 identical queries on different DB so no DB distinction is done during invalidation. If my understanding is correct, a way to keep in line with the current behavior is perhaps to use a default CACHEOPS_PREFIX which return empty string if a model is not db_agnostic and the db connection name if it is. If so, the db_agnostic clause in _cache_key method would be removed.

tests/tests.py
Svan70: I would not say they are unneeded. They test a precise point : that concerned methods call invalidators with good parameters. Of course the other will fail if these one fail but they are more 'unitary'. As you wish, I can remove them.

cacheops/query.py
Svan70: For post_save, instance._state.db is always equal to using.
https://github.com/django/django/blob/master/django/db/models/base.py#L754
(I work with django 1.9, it the same behavior).

PR #296, opened 2018-08-14
Call "unlink" to del conj sets if Redis > v4.0

nicwolff:
Redis 4.0 offers asynchronous garbage collection of deleted and expired keys and their values. We believe that deletion of large conjunction set values stored by cacheops is the cause of potential severe spikes in Redis latency that block all reads.

**This change updates django-cacheops to call UNLINK instead of DEL on conj sets if the Redis version is at least 4.0.**


Suor:
I would rather rewrite the script to not send the argument each time. And `invalidate_model()` should also be updated.

PR #298, opened 2018-08-30
Make call if cache result is none

junaid1460:


Suor:
This is checked at the start of the function.

PR #301, opened 2018-09-14
Fix test database to work in environments without /dev/shm

georgejlee:
Running the tests failed for me on MacOS without this change. I verified that the tests now work for me on MacOS with both Python 2.7 and Python 3.7.


Suor:
Thanks

PR #302, opened 2018-09-14
Add option to @cached_as to prevent caching on concurrent invalidation

georgejlee:
This change allows us to ensure that a @cached_as function always returns valid results and that stale data is not cached. Without this change, there are at least two possible race conditions:

### Case 1: Caching stale data

Consider a function `get_book_names()` decorated with `@cached_as(Book.objects.all())`. This function executes the queryset `Book.objects.all()` to return an array of book names. The cache needs to be invalidated if any book changes.

1. Process 1 calls `get_book_names()`. Its return value is not yet cached so Process 1 begins evaluating the function `get_book_names()` and executes the query to get all books.
2. Process 2 modifies an book `b` and changes its name from "Stale" to "New" and saves it.
3. All cache keys that depend on `b` are invalidated.
4. Process 1 finishes evaluating `get_book_names()`. This list of book names still has the name "Stale" for book `b`. This list is now cached.
5. Calling `get_book_names()` again returns the cached value with stale data.

In this case, the stale data should not be cached so that future requests will get the correct book names. It would be even better if this function call returned the new book names.

### Case 2: Returning inconsistent data

Consider a function `get_book_author_names()` decorated with `@cached_as(Book, Author)`. This function executes the queryset to return an array of books and their authors. The cache needs to be invalidated if any book or author changes.

1. Process 1 calls `get_book_author_names()`. Its return value is not yet cached so Process 1 begins evaluating the function `get_book_author_names()` and executes the query to get all books. Process 1 iterates through each book and for each book executes a separate query to get its author. Suppose there are 10 books by author "Sue". Process 1 iterates through the first five books by Sue to get the book and author names.
2. Process 2 modifies the author "Sue" and changes her name from "Sue" to "Susan" and saves it.
3. All cache keys that depend on Sue are invalidated.
4. Process 1 iterates through the remaining five books by Sue. These books now have author name "Susan". The resulting list of books now has five books by Sue and five by Susan, which is inconsistent.

This problem could be fixed by getting all books and authors in a single SQL query or by wrapping the entire call in a transaction, but the fix for Case 1 also addresses Case 2.

### The fix

We would like to enforce the invariant that if the samples that the function depends on are invalidated, that the function's return value is also invalid. To enforce this condition, we add an boolean option `retry_until_valid` to `@cached_as` that does the following:
* Before calling the function, we store a precache key unique to this function invocation that will get invalidated whenever the samples are invalidated.
* Before returning a value, check to see if the precache key is still present. If not, store the precache key again and retry calling the function.
* After caching the result, check to see if the precache key is still present. If not, invalidate the cached result and retry calling the function.

Note that if your function does any internal caching, this will need to be reset when the function is invoked to prevent stale data from persisting across retries. (e.g. call `copy.deepcopy()` on QuerySets to clear the result cache).

I added a `max_retry_count` parameter to help prevent infinite loops. I decided to raise an exception when the retry count is reached to ensure that invalid data is never returned. If it is acceptable to stale data you could change this behavior.

#### Automated tests
I added two automated tests:
1. `test_cached_as_retries_if_invalidated_during_func`
2. `test_cached_as_raises_exception_if_too_many_retries`

#### Manual tests
I verified the cases above in my web application with multiple worker processes.

### DIFF
--- a/cacheops/invalidation.py
+++ b/cacheops/invalidation.py
@@ -23,6 +23,13 @@ def redis_can_unlink():
     return StrictVersion(redis_version) >= StrictVersion('4.0')
 
 
+def invalidate_keys(*keys):
+    if redis_can_unlink():
+        redis_client.execute_command('UNLINK', *keys)
+    else:
+        redis_client.delete(*keys)
+
+
 @queue_when_in_transaction
 @handle_connection_failure
 def invalidate_dict(model, obj_dict, using=DEFAULT_DB_ALIAS):
@@ -63,10 +70,7 @@ def invalidate_model(model, using=DEFAULT_DB_ALIAS):
     if conjs_keys:
         cache_keys = redis_client.sunion(conjs_keys)
         keys = list(cache_keys) + conjs_keys
-        if redis_can_unlink():
-            redis_client.execute_command('UNLINK', *keys)
-        else:
-            redis_client.delete(*keys)
+        invalidate_keys(*keys)
     cache_invalidated.send(sender=model, obj_dict=None)
 
 
--- a/cacheops/lua/cache_thing.lua
+++ b/cacheops/lua/cache_thing.lua
@@ -1,9 +1,15 @@
 local prefix = KEYS[1]
 local key = KEYS[2]
+local precall_key = KEYS[3]
 local data = ARGV[1]
 local dnfs = cjson.decode(ARGV[2])
 local timeout = tonumber(ARGV[3])
 
+if precall_key ~= '' and not redis.call('get', precall_key) then
+  -- Cached data was invalidated during the function call. The data is
+  -- stale and should not be cached.
+  return
+end
 
 -- Write data to cache
 redis.call('setex', key, timeout, data)
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -123,7 +123,8 @@ def cached_as(*samples, **kwargs):
                     # the key to prevent falsely thinking the key was not
                     # invalidated when in fact it was invalidated and the
                     # function was called again in another process.
-                    precall_key = prefix + 'asp:' + key_func(func, args, kwargs, key_extra + [random()])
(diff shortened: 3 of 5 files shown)


Suor:
Issue 1 is not specific to `@cached_as()`. Any queryset can be theoretically invalidated right between db fetch and cache write, which will lead to caching stale data. However, as far as I know there were no real issues about that for anyone. Did you catch that in your project?

Issue 2 is not an issue of cacheops, this is expected behavior of relational databases and should not be fixed by cacheops but by means of dbs like transactions or reformulating queries.

georgejlee:
Yes, caching stale data was causing problems for us. It is true that Django’s queryset caches could also cache stale data, but those would be refreshed on the next API call. We are using cacheops as a more long lived cache for serialized responses. These serialized responses can take some time to generate from the raw DB queries and requests can be quite frequent so it is a huge performance boost to cache them. But the results must be accurate on the next API call.

Suor:
So you have a computation delay between fetch and cache write, which increases a probability of caching stale data. Did you try using `SELECT FOR UPDATE` at the start of cached function? This will prevent updates during call.

georgejlee:
Thanks for the suggestion. I'd like to avoid using `SELECT FOR UPDATE` to reduce the chance of deadlock and for improved performance. It also can be quite tricky to ensure the `SELECT FOR UPDATE` query is correct for the function call. The main problem I'd like to avoid is caching stale data for a prolonged time. What if I take out the retry logic so that it just invalidates the cached result if the data was invalidated during the function call?

Suor:
Is second check - that invalidation happened during caching - really needed?

Suor:
And anyway, it could be better to pass precall key into `cache_thing.lua` and just abort if that not exists.

georgejlee:
The second check protects against a race condition where the invalidation occurs after the previous check but before the cache write. This is unlikely, but still possible. I like the idea of doing the check in `cache_thing.lua`, that does seem cleaner. I'll look into that approach.

PR #303, opened 2018-09-17
Pass connection options through CACHEOPS_SENTINEL

### LINKED ISSUE #286: add ability to add more connection options to sentinel through CACHEOPS_SENTINEL
any other options set in CACHEOPS_SENTINEL other than those documented should be passed along directly to sentinel either in the constructor or master_for() call..  

currently there seems to be no way to ovverride the connection pool size or to override socket_connect_timeout for the actual redis client or even a way to socket_connect_timeout for the sentinel itself

davidfb:
fixes #286 This changes make possible to define socket timeouts both for Sentinel and Redis connections, as well as any other extra configuration accepted by the connection pool.

### DIFF
--- a/cacheops/redis.py
+++ b/cacheops/redis.py
@@ -95,7 +95,7 @@ def redis_client():
 
         sentinel = Sentinel(
             settings.CACHEOPS_SENTINEL['locations'],
-            omit(settings.CACHEOPS_SENTINEL, ('locations', 'service_name', 'db')))
+            **omit(settings.CACHEOPS_SENTINEL, ('locations', 'service_name', 'db')))
         return sentinel.master_for(
             settings.CACHEOPS_SENTINEL['service_name'],
             redis_class=client_class,
--- a/cacheops/redis.py
+++ b/cacheops/redis.py
@@ -6,7 +6,7 @@ import six
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.module_loading import import_string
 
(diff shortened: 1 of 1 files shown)


Suor:
So you are passing `socket_timeout` earlier now, is it the same?

Also, can you please use [omit()](https://funcy.readthedocs.io/en/latest/colls.html#omit) instead of dict comprehension?

davidfb:
Just added a commit to use `omit()` instead of dict comprehension.

About `socket_timeout`, all parameters passed to Sentinel initializer other than the Sentinel-specific ones (`sentinels`, `min_other_sentinels` and `sentinel_kwargs`) are passed along as `connection_kwargs` and will be used when establishing connections to the Redis server. This includes, among others, `socket_timeout` as well as `socket_connect_timeout`, `socket_keepalive` and `socket_keepalive_options`, that will be properly applied to all Redis connections. The only difference is that the default behaviour in redis-py is to also use all 4 of these parameters when stablishing the connection to Sentinel, so with this change, all socket-related options will start to apply to Sentinel connections as well (previous to this commit there was no way to configure timeouts for Sentinel). If specific configuration were to be used for Sentinel connections that diverges from the one to be used on connections to Redis instances, it would be enough to add a `sentinel_kwargs` parameter with that configuration in `settings.CACHEOPS_SENTINEL`. This mimics redis-py interface.

Suor:
Thanks

PR #304, opened 2018-10-03
Support for IDE like PyCharm

Tusky:
The previous code is not recognizable by PyCharm, with this small change it is and the end result should be the same once tests/docs are changed accordingly too.

If backwards compatibility is an issue, which it is definitely since it changes the @decorator_tag tag name, I created a an issue from PyCharm's end too: https://youtrack.jetbrains.com/issue/PY-32036

### DIFF
--- a/cacheops/templatetags/cacheops.py
+++ b/cacheops/templatetags/cacheops.py
@@ -33,9 +33,9 @@ __all__ = ['CacheopsLibrary', 'invalidate_fragment']
 
 
 class CacheopsLibrary(Library):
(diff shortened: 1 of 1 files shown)


Suor:
I don't like it. This is not inheritance and should not look like it. Can also broke someones code if he uses `CacheopsLibrary` instance as a drop-in. BTW, `@decorator_tag` is public API you can't just rename it.

So definitely no. Should be fixed from PyCharm side or any other way without breaking things.

PR #307, opened 2018-11-09
Add invalidate_objs

harrybiddle:
In my use case I bulk insert 3500 objects with a `no_invalidation` context and then do the invalidation after. Calling `invalidate_obj` on each object individually took 10.0 seconds for me locally. With this `invalidate_objs` function that time goes down to 0.4 seconds.

### DIFF
--- a/cacheops/invalidation.py
+++ b/cacheops/invalidation.py
@@ -14,7 +14,7 @@ from .signals import cache_invalidated
 from .transaction import queue_when_in_transaction
 
 
-__all__ = ('invalidate_obj', 'invalidate_model', 'invalidate_all', 'no_invalidation')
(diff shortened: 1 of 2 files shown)


Suor:
It's fast, but it doesn't work at all. You merge dicts effectively loosing info, the last one will just overwrite everything.

harrybiddle:
Hmm, I see what you are saying. Am I going to have to modify the lua script to do achieve this? Do you think there is still a speed gain to be had?

Suor:
Depends on your use case. Also, this won't be correct anyway since you need to invalidate on both old and new object state.

Suor:
You can try limiting key set to whatever you filter on and then dedup dicts.

harrybiddle:
I tried to figure out how to do this from the codebase, but it was much more involved than I realised. In the end for my needs it is simpler for me to use the Redis client directly rather than via django-cacheops. Thank you for your help in any case, closing this PR out.

PR #322, opened 2019-04-25
Make conj exclusion rules into Django settings

nicwolff:
Cacheops makes "conj keys" to track the cached resultsets that should be invalidated when a resource is changed. Those keys are supposed to include all the fields in the queries that got those resultsets, but Suor felt that `TextField`s would be too large, and that `in` lists with more than 8 values would be too large, so cacheops leaves those out.

But we use `TextField`s for all text columns, and many of our queries do `in` of a large list of IDs. As a result, most cached resultset get added to generic conj keys. 

This is exacerbating the invalidation-spike issue, because those conj sets get way more keys they should – at our site, some get millions! – and when they are finally invalidated Redis blocks while deleting them all, which stops our application dead.

**This change makes those conj exclusion rules into Django settings**, defaulting to their current values.

### DIFF
--- a/cacheops/conf.py
+++ b/cacheops/conf.py
@@ -23,6 +23,7 @@ class Defaults(namespace):
     CACHEOPS_DEGRADE_ON_FAILURE = False
     CACHEOPS_SENTINEL = {}
     CACHEOPS_NOT_SERIALIZED_FIELDS = models.FileField, models.TextField, models.BinaryField
+    CACHEOPS_LONG_DISJUNCTION = 8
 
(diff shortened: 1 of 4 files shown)


nicwolff:
Fails the same test `master` branch does.

Suor:
I see this, just don't have time to go about it. But since I am here:
- you removed explanation why those skipped
- if this goes public I would use some easier to understand name, like `CACHEOPS_SKIP_FIELDS`.

Suor:
Also, renamed `CACHEOPS_NON_SERIALIZED_FIELDS` to `CACHEOPS_SKIP_FIELDS`.

PR #323, opened 2019-05-02
Add conj-reaper management command

nicwolff:
We're running cacheops in a Django Rest Framework application that serves about 20,000 requests per minute, and our biggest conj sets – for which all filters were excluded from the DNF, and the table is rarely updated – had been growing to contain over 20 million keys.

On invalidation, this blocked Redis (and our entire application!) for many seconds while the `invalidate` `EVALSHA` looped over the keys in the conj set to delete query keys. (Even [using `UNLINK` for the deletion of the conj set itself](https://github.com/Suor/django-cacheops/pull/296) only helped halfway.)

So we've added this Django management command, which loops continually to scan cacheops' conj keys for already-expired cache keys and removes them. It can be configured with command-line options or environment variables to run as fast as your CPUs can accommodate, and to ignore conj sets under some reasonable size.

Some other users may find it valuable, so I submit it for inclusion with your excellent library.

### DIFF
--- a/cacheops/management/commands/conj-reaper.py
+++ b/cacheops/management/commands/conj-reaper.py
@@ -1,0 +1,86 @@
+import logging
+import time
+from os import getenv
+
+import redis
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+logger = logging.getLogger(__name__)
+log_tmpl = 'Removed %s/%s cache keys from %s'
+
+
+"""conj-reaper - A Django command to remove expired cache keys from invalidation sets.
+
+cacheops saves each DB resultset cache key in a "conj set" so it can delete it later if it thinks it
+should be invalidated due to a saved record with matching values. But the resultset caches time out
+after 30 minutes, and their cache keys live in those conj sets forever!
+
+So conj sets for frequent queries on tables that aren't updated often end up containing millions
+of already-expired cache keys and maybe a few thousand actually useful ones, and block Redis for
+multiple - or many - seconds when cacheops finally decides to invalidate them.
+
+This is a Django command that loops forever scanning cacheops' conj keys for already-expired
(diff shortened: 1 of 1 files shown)


Suor:
I suspect if you are using ENV here then that's what you are using in your app. There is, however, a major way in Django - use settings everywhere, if you need to set something from ENV then assign a setting from env var in `settings.py`:

```python
CACHEOPS_REAPER_SLEEP = os.environ.get('CACHEOPS_REAPER_SLEEP', 0.01)
```
There are many libs to support this like [django-environ](https://github.com/joke2k/django-environ) or [dj_database_url](https://github.com/kennethreitz/dj-database-url). This way every library doesn't need to deal with it. And you can use:

```
from cacheops.redis import redis_client

redis_client.sscan(...)
```
While using it with:
```bash
CACHEOPS_REDIS="redis://redis/7" CACHEOPS_REAPER_SLEEP=0.05 ./manage.py reap-conjs ...
```

bharling:
I'd prefer to see this implemented as a command in the library as with the `invalidate_foo` commands already available? or even just have it handled automatically without the developer needing to do anything.

if it was a library-level command it could easily be dropped into a periodic celery task or something?

codeasashu:
I have memory issues with this library and I badly need this merged. Do we have any timeline for this?

Suor:
> if it was a library-level command it could easily be dropped into a periodic celery task or something?

When it needs to be implemented as a celery task. Or maybe a function, which would be easy to wrap into celery or any other bg task. Anyway not a function.

> I have memory issues with this library and I badly need this merged. Do we have any timeline for this?

For this to be merged it needs to be cleaned up. It might be used as is though.

Suor:
Implemented in #434.

PR #334, opened 2019-10-02
Added custom manager and queryset support

### LINKED ISSUE #333: Unable to use custom managers or querysets
A common practice is to create a manager to do "soft" deletes on data when that data needs to be tracked.  Unfortunately instead of deriving from `models.Manager`, the manager I have to deal with looks like:

```
class SoftDeleteManager(manager.BaseManager.from_queryset(SoftDeleteQuerySet)):
    use_for_related_fields = True

    def get_queryset(self):
        return super(SoftDeleteManager, self).get_queryset().filter(active=True)
```
So it would be nice to be able to set the custom manager class that gets monkey patched.  Currently it has to derive from `models.Manager`, but it doesn't have too to be a valid manager.

Pull request is incoming

xuru:
Fixes #333 for custom managers
- Updated check for manager to support custom managers
- Added tests
- Added exception handling around importing custom managers

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,15 @@
+4.2
+- support Django 2.1 and 2.2
+- added keep_fresh option to @cached_as (George Lee)
+- pass CACHEOPS_SENTINEL options through (David Fuentes Baldomir)
+- made SKIP_FIELDS and LONG_DISJUCTION configurable (Nic Wolff)
+- fixed .aggregate() over .annotate() fields in Django 2.2
+- fixed .bulk_create() in Django 2.2 (Grzegorz Szczepanczyk)
+- fixed test database to work in environments without /dev/shm (George Lee)
+- fixed proxy/abstract model bug
+- added test case for issue #312 (Andy Tzeng)
+- some warnings and doc-strings improvements
+
 4.1
 - use UNLINK instead of DEL when available, use Redis 4.0+ for that
 - request m2m objects for invalidation using correct db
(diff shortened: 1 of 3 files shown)


Suor:
As far as I see from Django code, all managers descend from `BaseManager`, so cacheops might just switch to patch that instead of `Manager`. No extra code will be required.

Suor:
Also test could be simplified - a noop custom QuerySet and anonymous Manager should do. Also shouldn't test internals in test body.

xuru:
This totally makes sense.  Kind of over engineered it.  I'll make the changes, and see how it works.

xuru:
@Suor Ok, ready for your review, and thanks for the pointers.

cacheops/query.py
```diff
@@ -4,6 +4,8 @@
 import threading
 import six
 from random import random
+
+from django.db.models.manager import BaseManager
```
Suor: Please place this right after `from django.db.models import Model` import.

tests/tests.py
```diff
@@ -607,6 +606,12 @@ def test_312(self):
     def test_316(self):
         Category.objects.cache().annotate(num=Count('posts')).aggregate(total=Sum('num'))
 
+    def test_333(self):
+        # ensure that we can use a manager not derived from Manger
+        # (i.e. using manager.BaseManager.from_queryset)
+        mgr = CustomFromQSModel.objects
+        assert isinstance(mgr, CustomFromQSManager)
```
Suor: This doesn't really test anything. The old code will blow up in `install_cacheops()` call I suppose. You may just drop it.

cacheops/query.py
xuru: Done

PR #341, opened 2019-11-12
Disabled postponed invalidation when no_invalidation is applied

zerc:
In the current implementation when one calls `obj.save()` the `invalidate_obj` is [called](https://github.com/Suor/django-cacheops/blob/03735b99b48e290f8f4340bb599a1122e4392c82/cacheops/query.py#L483) regardless either we are inside of `no_invalidation` block or not.

`invalidate_obj` in turn calls `invalidate_dict` which is [marked](https://github.com/Suor/django-cacheops/blob/master/cacheops/invalidation.py#L32) by `queue_when_in_transaction` decorator which essentially pospones the invalidation until the transaction ends.

By the time it happens, we might be outside of the `no_invalidation` block so this [condition](https://github.com/Suor/django-cacheops/blob/master/cacheops/invalidation.py#L35) won't work and the cache invalidation will be performed which is, unexpected.

Given that Django may automatically wrap each request into a transaction via `ATOMIC_REQUESTS` setting it might be difficult to avoid the invalidation for such views.

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -473,7 +473,7 @@ class ManagerMixin(object):
                 pass
 
     def _post_save(self, sender, instance, using, **kwargs):
-        if not settings.CACHEOPS_ENABLED:
+        if not settings.CACHEOPS_ENABLED or no_invalidation.active:
             return
 
         # Invoke invalidations for both old and new versions of saved object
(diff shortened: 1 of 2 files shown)


Suor:
Great catch, Vladimir!

I'll take a closer look at this later today or tomorrow.

>

PR #352, opened 2020-04-30
Fix stamp_fields utility function

lokhman:
In `stamp_fields` utility function we create a stamp from the list of `(f.name, f.attname, f.db_column, f.__class__)` for all model fields returned from meta:

```python
stamp = str(sorted((f.name, f.attname, f.db_column, f.__class__) for f in model._meta.fields))
```

The last element is the actual class type of the field and not a class name. This will fail, as before being stringified the list is sorted, and the classes cannot be compared. We should use a class name instead.

Also, if `db_column` was not provided in model definition, `f.db_column` will be `None`, which is not comparable as well.

Practically, a model may have two fields with the same name and `attname` if one field was contributed to model as private. `model._meta.fields` return both local and private fields.

### DIFF
--- a/cacheops/utils.py
+++ b/cacheops/utils.py
@@ -65,7 +65,7 @@ def stamp_fields(model):
     Returns serialized description of model fields.
     """
     stamp = str(sorted(
-        (f.name, f.attname, f.db_column, f.__class__.__name__)
+        (f.name, f.attname, f.db_column or '', f.__class__.__name__)
         for f in model._meta.fields))
     return md5hex(stamp)
 
--- a/cacheops/utils.py
+++ b/cacheops/utils.py
@@ -65,7 +65,7 @@ def stamp_fields(model):
     Returns serialized description of model fields.
     """
     stamp = str(sorted(
-        (f.name, f.attname, f.db_column or '', f.__class__.__name__)
+        json.dumps(f.deconstruct(), sort_keys=True, default=obj_key)
         for f in model._meta.fields))
     return md5hex(stamp)
 
--- a/cacheops/utils.py
+++ b/cacheops/utils.py
@@ -64,7 +64,9 @@ def stamp_fields(model):
     """
     Returns serialized description of model fields.
     """
-    stamp = str(sorted((f.name, f.attname, f.db_column, f.__class__) for f in model._meta.fields))
+    stamp = str(sorted(
+        (f.name, f.attname, f.db_column, f.__class__.__name__)
+        for f in model._meta.fields))
     return md5hex(stamp)
 
 
--- a/tests/models.py
+++ b/tests/models.py
@@ -226,14 +226,31 @@ class Device(models.Model):
 
 
 # 333
-class CustomeQuerySet(QuerySet):
+class CustomQuerySet(QuerySet):
     pass
 
 
-class CustomFromQSManager(manager.BaseManager.from_queryset(CustomeQuerySet)):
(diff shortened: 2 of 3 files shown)


Suor:
Can you provide a failing test?

Suor:
As far as I see only names are compared, so there is no issue.

lokhman:
Here is a completely legitimate case, when two fields are added with the same name:
```python
class CombinedCharField(models.CharField):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

        self.field = models.CharField(*args, **kwargs)

    def contribute_to_class(self, cls, name, **kwargs):
        super().contribute_to_class(cls, name, private_only=True)

        self.field.contribute_to_class(cls, name, **kwargs)
    ...

class MyModel(models.Model):
    my_field = CombinedCharField(max_length=8)
```

If cacheops is enabled for `MyModel` you will not be able to run any manager methods (e.g. `MyModel.objects.all()`), as it will be failing on `stamp_fields` firstly trying to sort by two `db_column=None` and then (if you specify `db_column` manually) by comparing field classes, that are not comparable.

Suor:
Ok, then we need both class name and module to be safe. Or probably we should use `Field.deconstruct()` instead of our hand-made tuole. Also still need a test.

lokhman:
Yes, `Field.deconstruct()` would be better in this case.

In what test case do you see the test to be added? I assume something like that should be fine?
```python
class MyModel(models.Model):
    my_name = models.CharField(max_length=8)

fields = list(MyModel._meta.fields)
new_field = models.BooleanField(name='my_name')
new_field.set_attributes_from_name('my_name')
fields.append(new_field)
MyModel._meta.fields = fields

stamp = cacheops.utils.stamp_fields(MyModel)
self.assertEqual('6b0a8554bcb827ea1eeb2975b2522a3c', stamp)
```

Suor:
You shouldn't test particular stamp, this might change on Django update. Test should fail the way your code does.

lokhman:
If we use field name and field module/class (the first two items of deconstruct) the stamp should always be the same unless Django changes the way how the field path is created. What do you suggest?

Suor:
I think we should use the whole deconstruct thing.

lokhman:
The last deconstruct item (kwargs) is not comparable as well, and the third (args) can potentially be not comparable also. We can add `db_column` only but I don't think of a use case when two fields have the same names and types.

Suor:
Can do `json.dumps(field.deconstruct, default=str)`. Use cases are many, say someone changes `JSONField(encoder)` and the new one transforms data somehow upon fetch, cache should be invalidated.

lokhman:
Should probably be better with `json.dumps(f.deconstruct(), default=str, sort_keys=True)`. You reckon it won't hit the performance?

Suor:
No, the function is memoized.

Suor:
Should probably use `default=obj_key` same as in `func_cache_key()`, this will prevent memory addresses in function str, some might be used as defaults.

lokhman:
Makes sense. Added code and the test.

lokhman:
Makes sense. Added code and the test.

Suor:
Looks good, thank you.

lokhman:
@Suor do you plan to make any new releases soon?

lokhman:
Cool, we'll wait for your release to switch from fork then. Thanks!

Suor:
Released in cacheops 5.0.

PR #370, opened 2020-08-10
Extend dnfs() to catch subqueries used for annotation

### LINKED ISSUE #365: Feasibility of supporting Subquery invalidation
I understand that there's a known limitation concerning the invalidation of subqueries. The README says:

> Conditions on subqueries don't affect invalidation. ... [This] can be implemented if needed, but it's probably counter-productive since one can just break queries into simpler ones, which cache better.

I'm curious what degree of effort that would entail, as I recently discovered a use case for it under Django 3.1. I'll do my best to convey my current dilemma.

### Current Implementation

Assume two related models, Blog and Article:

```python
class Blog(models.Model):
    name = models.CharField()

    class Meta:
        ordering = ['name']


class Article(models.Model):
    blog = models.ForeignKey(to=Blog, related_name='articles')
    name = models.CharField()
    body = models.TextField()
```

When fetching a list of blogs, we annotate the number of articles associated with each using `Count()`. For example:

```python
blogs = Blog.objects.annotate(article_count=Count('articles'))
```

This works. Caching works. Everything works.

### Django 3.1

Django 3.1 [now removes the default ordering](https://docs.djangoproject.com/en/3.0/topics/db/aggregation/#aggregation-ordering-interaction) from queries when a `GROUP BY` clause is present, such as when using `Count()` above. This means that the above queryset still works, but the list of blogs returned is no longer ordered by name.

The presumed solution is to attach `.order_by('name')` to the queryset to restore the ordering. And indeed this works, however it's obviously not ideal to have to attach this to _every_ affected query for _every_ model in an application. Distaste for boilerplate aside, it's very easy to forget here and there.

As a workaround, I've experimented with using [subqueries](https://docs.djangoproject.com/en/3.0/ref/models/expressions/#subquery-expressions). It's a bit unwieldy, but the following works (and actually seems to be more efficient):

```python
article_count = Article.objects.filter(blog=OuterRef('pk')).values('blog').annotate(c=Count('*')).values('c')
blogs = Blog.objects.annotate(article_count=Subquery(article_count))
```

This works, and blogs are ordered, but of course now we've run into the issue with subqueries. Namely, once the above query has been cached, creating a new article on a blog won't update its count until the cached result expires.

All this is to ask: How feasible would it be to support the above use case? Any insight you can provide is greatly appreciated.

jeremystretch:
Closes #365: Support invalidation for annotation subqueries

Extends `dnfs()` to check for subqueries used within `annotate()` on a queryset and, if present, attach them to the DNF tree. This is my first foray into the inner workings of cacheops, so feedback is much appreciated!

### DIFF
--- a/cacheops/tree.py
+++ b/cacheops/tree.py
@@ -129,7 +129,9 @@ def dnfs(qs):
     # Add any subqueries used for annotation
     if qs.query.annotations:
         subqueries = (
-            query_dnf(q.query) for q in qs.query.annotations.values() if type(q) is Subquery
+            # Django 3.0+ sets Subquery.query
+            query_dnf(getattr(q, 'query', q.queryset.query)) for q in qs.query.annotations.values()
+            if type(q) is Subquery
         )
         dnfs_.update(join_with(lcat, subqueries))
 
--- a/cacheops/tree.py
+++ b/cacheops/tree.py
@@ -1,6 +1,7 @@
 from itertools import product
 from funcy import group_by, join_with, lcat, lmap
 
+from django.db.models import Subquery
 from django.db.models.query import QuerySet
 from django.db.models.sql import OR
 from django.db.models.sql.query import Query, ExtraWhere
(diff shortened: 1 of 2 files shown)


Suor:
Thanks. A few things:

1. Do we ignore subqueries outside annotations?
2. Need a test.
3. No need to implement it in two stages. We may collect all queries first.

jeremystretch:
> 1. Do we ignore subqueries outside annotations?

Yes. This PR is limited to addressing the use case I describe in #365. Hopefully others can contribute similar additions as other use cases are identified.

> 3. No need to implement it in two stages. We may collect all queries first.

Sorry, I don't follow. Can you elaborate?

Suor:
You can collect queries:
```python
queries = ...

if self.query.annotations:
    queries += ...
```
Then process all of them. Also, some of the combined queries might also contain annotations.

jeremystretch:
I'm afraid what I've submitted so far is about the extent of what I'll be able to contribute. If that's sufficient, I'm happy to add a test before it gets merged. However if the project needs a more complete implementation to proceed, unfortunately I won't be able to make that commitment.

Suor:
A test would be great.

jeremystretch:
Added a test and fixed compatibility for Django <3.0

Suor:
Thanks, merged.

lampwins:
@Suor would it be possible to get a new release with this included?

Suor:
I'll take a look in a couple of days. I might want to add more fixes though.

Suor:
Released in cacheops 5.1.

PR #376, opened 2020-09-07
MANIFEST.in: Update

jayvdb:


Suor:
Why?

jayvdb:
Because distro packaging requires tests.

Suor:
What do you mean by distro packaging and why it requires tests?

Suor:
Also why tests can't be taken from here instead of pypi?

jayvdb:
I do packaging for openSUSE.
pip defaults to using wheels if you are providing them.  If not, you should be, as they are safer, smaller, quicker, etc,etc .
packagers use sdists if possible.
Getting the tarballs from GitHub is extra work for me, every release, and you would believe how often GitHub tags have different content from what is on PyPI, or maintainers stop using tags.
Anyawy, if you dont want this in distros, close it and the issue.

Suor:
Ok, makes sense now.

PR #396, opened 2021-04-16
Adds prefix to all keys.  Adds ignore_prefix kwarg.

### LINKED ISSUE #325: cacheops.cache.set/get/delete does not include the CACHEOPS_PREFIX on the keys
Hello, is it expected that the [`cacheops.cache`](https://github.com/Suor/django-cacheops/blob/883cc9448d9954bb83257f21cf20c586135f2e06/cacheops/simple.py#L81-L103) methods don't automatically include on the key the value returned by `CACHEOPS_PREFIX` if that function/value was defined?

if yes (that is, the current behavior is expected), maybe this should be documented with a Big Warning.

Anyway, thanks for the incredible work at this library.

scuml:
Resolves #325 

Adds the prefix to all decorators, and simple caching methods.  As this plugs a giant potential security issue where cache can leak between different deployments and applications, namespacing all keys is critical and should be the default behavior.  I've included the ability to revert to the previous key names by adding the parameter `ignore_prefix` to the decorators and simple cache methods.  This allows cache to be shared by eliminating the prefix, but only if the user explicitly requests it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -650,7 +650,20 @@ A ``query`` object passed to callback also enables reflection on used databases 
         if query.tables == ['blog_post']:
             return 'blog:'
 
-**NOTE:** prefix is not used in simple and file cache. This might change in future cacheops.
+**NOTE:** Prefix is now used on all cacheops operations.  This was not the case for simple and file cache in cacheops versions 5.1 and earlier.  To share cache between multiple applciations, cache can be saved without the prefix by including ``ignore_prefix=True`` to any of the decorators:
+
+.. code:: python
+
+    from cacheops import cached, cached_view
+
+    @cached(timeout=number_of_seconds, ignore_prefix=True)
+    def top_articles(category):
+        return ... # Some costly queries
+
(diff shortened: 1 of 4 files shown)


Suor:
Thanks for your interest @scuml.

If you want to add prefix to simple cache then you should do it at `BaseCache` level. You will probably need to provide `.get()`, `.set()` and `.delete()` there calling backend specific `._get()` and friends from its descendants, this will ensure you won't have discrepancy between them.

On implementation - let's not add `ignore_prefix` so far. Switching back on function by function basis doesn't really look practical anyway. We may think about how it can be exposed on `PrefixQuery` though, so that people will be able to opt out on config level. Don't need to go into this PR.

Also please avoid whitespace only changes, i.e. new empty newlines.

Suor:
Closing in favor of https://github.com/Suor/django-cacheops/commit/e29dcbc8fbf1845e9249bd5a4a8462fa846afb80.

PR #399, opened 2021-04-29
Fix inconsistency of `transaction_states`

ron8mcr:
Original case was:
- Django with `ATOMIC_REQUESTS` enabled
- Cache stopped work randomly on random `uwsgi` workers (while other continued to use/invalidate cache)
- Cache started to work after restarting worker
    
The problem was that neither `TransactionState.commit`, nor `TransactionState.rollback` code was called after `IntegrityError`, so `transaction_states` kept one not handled `TransactionState` and marked all other requests as `dirty` and didn't use cache.

### DIFF
--- a/cacheops/transaction.py
+++ b/cacheops/transaction.py
@@ -77,7 +77,7 @@ class AtomicMixIn(object):
             self._no_monkey.__exit__(self, exc_type, exc_value, traceback)
         except DatabaseError:
             transaction_states[self.using].rollback()
-        finally:
+        else:
             if not connection.closed_in_transaction and exc_type is None and \
                     not connection.needs_rollback:
                 if transaction_states[self.using]:
--- a/cacheops/transaction.py
+++ b/cacheops/transaction.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 
 from funcy import once, decorator
 
-from django.db import DEFAULT_DB_ALIAS
+from django.db import DEFAULT_DB_ALIAS, DatabaseError
 from django.db.backends.utils import CursorWrapper
 from django.db.transaction import Atomic, get_connection, on_commit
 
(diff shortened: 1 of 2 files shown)


cacheops/transaction.py
```diff
@@ -73,13 +73,17 @@ def __enter__(self):
...
+            self._no_monkey.__exit__(self, exc_type, exc_value, traceback)
+        except DatabaseError:
             transaction_states[self.using].rollback()
+        finally:
+            if not connection.closed_in_transaction and exc_type is None and \
+                    not connection.needs_rollback:
+                if transaction_states[self.using]:
+                    transaction_states[self.using].commit()
+            else:
+                transaction_states[self.using].rollback()
```
Suor, line 86: It looks like rollback might be called twice. Also, both commit and rollback might be called. Changing `finally` to `else` will fix both of that.

tests/tests_transactions.py
```diff
@@ -91,6 +91,22 @@ def test_smart_transactions(self):
...
 
+    def test_rollback_during_integrity_error(self):
+        # store category in cache
+        get_category()
+
+        # Make current DB be "dirty" by write
+        try:
+            with atomic():
+                Post.objects.create(category_id=-1, title='')
+        except (IntegrityError, Exception):
```
Suor, line 102: `Exception` includes `IntegrityError`. So which on is here?

Suor (changes requested):

Suor:
Thanks for PR @ron8mcr, can you please look at my comments above?

tests/tests_transactions.py
ron8mcr: `Exception` was added for debugging test, and left accidentally. Removed it

cacheops/transaction.py
ron8mcr: Thanks, fixed

Suor:
Thanks, merged.

PR #402, opened 2021-05-20
Add cluster support

longhoangpp:
- In this PR, we have added the support for cluster mode to django-cacheops
- The feature is toggle-able via `CACHEOPS_CLUSTER_ENABLED` flag
- In order to make it works, we moved the logic from the Lua scripts to Python with the support of RedisCluster client from `redis-py-cluster` package
- We also created tests (inherited from normal tests) for cluster mode, people can run it by running `run_tests_cluster.py` file
- While we were implementing this mode, we wanted all the keys will be distributed evenly. That's why we added a prefix check in cluster mode to prevent user adding a redis hashtag to the prefix, which will lead to unbalanced key distribution.

### DIFF
--- a/cacheops/redis.py
+++ b/cacheops/redis.py
@@ -1,4 +1,6 @@
 import warnings
+from funcy.decorators import wraps
+
 from contextlib import contextmanager
 
 from django.core.exceptions import ImproperlyConfigured
@@ -6,6 +8,7 @@ from django.utils.module_loading import import_string
 
 from funcy import decorator, identity, memoize, omit, LazyObject
 import redis
+from rediscluster import RedisCluster
 from redis.sentinel import Sentinel
 from .conf import settings
 
@@ -23,11 +26,28 @@ else:
     handle_connection_failure = identity
 
 
+def handle_timeout_exception(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except redis.TimeoutError as e:
+            error_handler = settings.CACHEOPS_TIMEOUT_HANDLER
+            if callable(error_handler):
+                error_handler(e, *args, **kwargs)
+
+            return None
+
+    return wrapper
+
+
 LOCK_TIMEOUT = 60
 
 
 class CacheopsRedis(redis.StrictRedis):
     get = handle_connection_failure(redis.StrictRedis.get)
+    # every function call this one, so better handle timeout at this level
+    execute_command = handle_timeout_exception(redis.StrictRedis.execute_command)
 
     @contextmanager
     def getting(self, key, lock=False):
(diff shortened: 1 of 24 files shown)


cacheops/redis.py
```diff
@@ -100,11 +177,22 @@ def redis_client():
...
     if isinstance(settings.CACHEOPS_REDIS, str):
-        return client_class.from_url(settings.CACHEOPS_REDIS)
+        redis_client = client_class.from_url(settings.CACHEOPS_REDIS)
+        redis_client.socket_timeout = settings.CACHEOPS_REDIS_CONNECTION_TIMEOUT
+        redis_client.socket_connect_timeout = settings.CACHEOPS_REDIS_CONNECTION_TIMEOUT
     else:
-        return client_class(**settings.CACHEOPS_REDIS)
+        kargs = {
+            "socket_timeout": settings.CACHEOPS_REDIS_CONNECTION_TIMEOUT,
+            "socket_connect_timeout": settings.CACHEOPS_REDIS_CONNECTION_TIMEOUT,
```
Suor, line 190: These might be just added to `CACHEOPS_REDIS`, no need for a separate setting.

Suor:
Hello, @longhoangpp, and thanks for using cacheops. I see that you obviously put an amount of effort in this, but I can't merge it because it will make cacheops unmaintainable for me. The reasons behind are:
- lots of code duplication as well as logic duplication
- races and possible partial failures in both cache thing and invalidation procedures will very soon lead orphan cache keys and then to stale cache
- storing cache keys and invalidators on separate nodes will lead to orphan keys in case of cluster split. The idea was to use `{prefix}` for sharding, which will ensure dependent keys will end in a single node. If one doesn't want that he is free not to use `{}`.
- your implementation is slow, e.g. reading schemas, reading conj sets, deleting all the stuff for invalidation will require lots of data going back and forth, cache thing procedure makes lots off independent calls. 
- you lock/unlock code is not atomic anymore, which means it probably won't work

First one makes it harder to maintain, the rest lead to huge number of complaints. I wish this still will work out for your use case though.

Some points on implementation.
- there is no need to get rid of all lua scripts, as long as their key go into the same node it should work. E.g. locking might be kept unchanged if `key` contains `{}` or otherwise simply change `signal_key = '{%s}:signal' % key`
- if you are ok with sharding rather than fully random distribution then using prefix like `{sharding_key}` and replacing `Redis` instance with `RedisCluster` might do the job for you
- for fully random key distribution you need to consider splits and partial writes/invalidations. The one way I can think of is flipping responsibility on invalidation check - save timestamps rather than sets in conj keys, save timestamp with cache data too, on each read `MGET` cache_key and all its conj keys, check that conj keys are present and not updated later. On invalidation simply drop conj keys.
- if you are still going to implement part of things in Python you may look at older cacheops version before I moved to Lua, i.e. at how schemas are managed in local memory.

Hope this helps.

longhoangpp:
Thank you very much for your comment. We will consider your comments and fix it one by one, to improve our fork first, then maybe in near future, a good merge request to contribute our code to your code base.

Thank you once again

PR #405, opened 2021-06-03
Test related to issue 404

ymguerra:
This is the test for issue 404, this line it doesn't work '@override_settings(CACHEOPS = {'tests.post': {'ops': {'get', 'fetch'}},})' to be able to make it work I changed in my local the global settings  to
```CACHEOPS = {'tests.post': {'ops': {'get', 'fetch'}},}```
But I didn't push it, because I know it should be another way to do it.
Also here my test pass, so this is even more confuse for me

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -659,7 +659,26 @@ class IssueTests(BaseTestCase):
     def test_387(self):
         post = Post.objects.defer("visible").last()
         post.delete()
+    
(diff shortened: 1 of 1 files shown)


Suor:
It's ok to use `.cache()` if that demonstrates the thing. The fix won't depend on that.

Suor:
> Also here my test pass, so this is even more confuse for me

There should be something else to your code or setup.

PR #411, opened 2021-10-13
Add support to python 3.10

jairhenrique:


Suor:
Why do you remove Python 3.5? Also Travis doesn't work anymore.

PR #430, opened 2022-06-02
Fix `invalidate_m2o` for polymorphic models

aletor123:
Now there is problem with polymorphic models and `invalidate_m2o`.
When Model is inherited it's pk field is `OneToOne` field to parent model. And if there is some other model with FK to inherited model, this FK points to `OneToOne` field. In this situation `getattr(instance, f.field_name)` will return model instance of parent model, when we need to get `id`.

We can fix it using Model._meta.pk.attname.
Test case presented.
Can you please add this fix to 6.1

### DIFF
--- a/cacheops/query.py
+++ b/cacheops/query.py
@@ -529,10 +529,12 @@ def invalidate_m2o(sender, instance, using=DEFAULT_DB_ALIAS):
     """Invoke invalidation for m2o and m2m queries to a deleted instance"""
     all_fields = sender._meta.get_fields(include_hidden=True, include_parents=True)
     m2o_fields = [f for f in all_fields if isinstance(f, models.ManyToOneRel)]
+    fk_fields_names_map = {
+        f.name: f.attname
+        for f in all_fields if isinstance(f, models.ForeignKey)
+    }
     for f in m2o_fields:
-        attr = f.field_name
-        if f.field_name == sender._meta.pk.name:
-            attr = sender._meta.pk.attname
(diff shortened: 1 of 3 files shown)


Suor:
First of all, welcome to cacheops and thanks for the PR. 

You should note though that multitable inheritance was never really supported by cacheops, in a sense that invalidation is not guaranteed but things should not blow up. Do they? It's not clear. Anyway, even if they don't I am willing to accept patches to improve multitable inheritance support as long as it's not me supporting them :), so we can proceed on that base too.

Some specifics:
- can you use existing models for tests instead of creating new ones? E.g. `Media` and `Movie`
- it is better to have a real test with no mocks and no internal calls like `invalidate_m2o()`, for now I don't understand what is that is broken and also see this test as fragile and hard to support one
- maybe the problem is that `OneToOneRel` is a subclass of `ManyToOneRel`? I.e. things may start to work if we restrict `invalidate_m2o()` strictly to `ManyToOneRel`. Or alternatively we simply need different handling for these two

Current code with `if` looks awkward and makes me suspect that it is not covering everything, but maybe I am wrong.

aletor123:
@Suor can you please check my description

aletor123:
Yes, you are right. I've updated implementation

Suor:
Thanks, I looked into it and it seems correct. I'll give it some more thought, the code doesn't look intuitive so may be missing something here. Although the previous code wasn't intuitive either, simply a bit less complicated )

PR #434, opened 2022-07-08
Add command to clear stale cacheops keys

browniebroke:
Building on top of the excellent work from @nicwolff  in #323. 

I found the code very useful and _almost_ good for us. The main problems we had were:

- we didn't want run this in a `while True` loop, we'd rather run it once a day or so
- we wanted it in Celery task. While it's probably not something to be implemented in cacheops, but more in our application, having the logic in a command was limiting its reusability, so I moved it into a function
- the admin command seems quite useful hence I left it

Note that I left the first commit to credit the original author.

I'm willing to put the extra steps to get this in cacheops. Is it something that you'd accept @Suor? If yes, let me know if it's good enough or what's missing. I guess tests & documentation would be nice 😄 Before I invest too much time, are you happy with how things are called?


cacheops/cleaner.py
```diff
@@ -0,0 +1,58 @@
...
+    So conj sets for frequent queries on tables that aren't updated often end up containing
+    millions of already-expired cache keys and maybe a few thousand actually useful ones,
+    and block Redis for multiple - or many - seconds when cacheops finally decides
+    to invalidate them.
+
+    This function scans cacheops' conj keys for already-expired cache keys and removes them.
+    """
+    logger.info('Starting scan for large conj sets')
+    prefix = get_prefix(dbs=['default'])
+    for conjunction_key in redis_client.scan_iter(f'{prefix}conj:*', count=chunk_size):
```
Suor, line 26: It's better to call it `conj_key` since this is what we have everywhere. Besides esthetic purposes it has a practical one too: imagine one searching through code for `conjuction` and for `conj`.

cacheops/cleaner.py
```diff
@@ -0,0 +1,58 @@
+import logging
```
Suor, line 1: Let's give a tribute to @nicwolff but saving his name `reaper`. Also, I simply like distinctive names)

cacheops/cleaner.py
```diff
@@ -0,0 +1,58 @@
...
+
+from .redis import redis_client
+from .sharding import get_prefix
+
+logger = logging.getLogger(__name__)
+
+
+def clear_stale_cacheops_keys(chunk_size: int, min_conj_set_size: int, dry_run: bool = False):
```
Suor, line 9: So this one will be `reap_conjs()` or smth. Also may move default here from the command.

Suor (approved):
Thanks for taking a stab at this. I like the approach it makes it more reusable. One might still want `while True: ... sleep()` loop though)

So docs will help a lot and some naming I would like to be fitted.

cacheops/cleaner.py
browniebroke: I think I might have to duplicate the default: if I remove the default from the command and the user doesn't specify them, then the logic will be called with `None`, which will cause an error.

cacheops/reaper.py
```diff
@@ -0,0 +1,58 @@
...
+
+    So conj sets for frequent queries on tables that aren't updated often end up containing
+    millions of already-expired cache keys and maybe a few thousand actually useful ones,
+    and block Redis for multiple - or many - seconds when cacheops finally decides
+    to invalidate them.
+
+    This function scans cacheops' conj keys for already-expired cache keys and removes them.
+    """
+    logger.info('Starting scan for large conj sets')
+    prefix = get_prefix(dbs=['default'])
```
browniebroke, line 25: That was hardcoded in our codebase to `'default'` but it needs changing. Should probably make it an argument of this function (like the `invalidate_...` functions)

Suor:
Thanks for the changes, looks good.

browniebroke:
Hello 👋🏻 

Sorry to revive this PR, I wanted to check if we can expect this to appear in a release relatively soon? If the release is around the corner, I can wait a bit more for it. On the other hand, I totally get it if you don't think you'll have time to do anytime soon or if there are other blockers, I can just install from the GitHub commit.

Thanks again for this great lib!

Suor:
I am preparing release now, probably, make it in a couple of days.

nicwolff:
Oh hey, I just saw this! Thanks! One more reason to update our cacheops install ツ

PR #440, opened 2022-11-22
Remove not needed six dependency

danigm:
The six package is not required anymore, python2 support was removed in https://github.com/Suor/django-cacheops/commit/19c36b2bb51aa05188ee8f6ad23cac1f850d7c66


Suor:
Thanks

PR #443, opened 2023-01-12
fix abstract models that do not have an application

coldcloudgold:
Solving the problem when trying to handle the cache for abstract models that do not have an application (for third-party django applications)


Suor:
Why cache abstract models that do have an app?

Suor:
BTW, also, why do you try to cache them in the first place? I.e. why do you add them to `CACHEOPS` settings?

Suor:
Fixed in 55179e3657df351257c54faa07af1bd8c39e3a0c

PR #453, opened 2023-05-05
More lenient way of checking request object

modbender:
More generic way of checking request object

Source: Django view decorator cache.py
https://github.com/django/django/blob/main/django/views/decorators/cache.py

### DIFF
--- a/cacheops/utils.py
+++ b/cacheops/utils.py
@@ -4,7 +4,6 @@ import inspect
 from funcy import memoize, compose, wraps, any, any_fn, select_values, lmapcat
 
 from django.db import models
-from django.http import HttpRequest
 
 from .conf import model_profile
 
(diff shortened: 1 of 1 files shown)


Suor:
This looks like a very indirect way to check for type. Going duck typing helps though, I should agree. 

BTW, what about the other error you were getting?

modbender:
Working on that, but it will take time. This is just the first necessary step.

modbender:
For the other error, I'm planning on making a plugin package or something that can be used for rest API.
The classes used and working in REST API is way too different.

modbender:
I think this pull can be merged, it's harmless enough and might help someone out there.

Suor:
It is not as innocent as it looks:
1. It is incomplete, so it will make false promises of working and confuse people.
2. It has no tests, so might break silently at some point.

Merging this will require me to support it, which I am not ready to do in this form. Also, ideally it does not belong to cacheops as cacheops better not know about internals of other Django libs.

modbender:
Well the code is directly from official Django source code, so I don't think 1 is true.
About 2, I'm not all that knowledgeable about create tests yet, so I'll leave it to someone who might know or maybe do it later.

PR #455, opened 2023-05-16
Don't WRONGTYPE when activating CACHEOPS_INSIDEOUT with cached conj sets

nicwolff:
`CACHEOPS_INSIDEOUT` works by storing a random number for the `conj:` key and prefixing a checksum of those numbers to each stored resultset.

But without `CACHEOPS_INSIDEOUT` `conj:` keys have set values.

So if the application's cache already contains conj sets when `CACHEOPS_INSIDEOUT` is first activated, [the `cache_thing_insideout.lua` script](https://github.com/Suor/django-cacheops/blob/7.0.1/cacheops/lua/cache_thing_insideout.lua) will fail with
```
WRONGTYPE Operation against a key holding the wrong kind of value script: 4dfdfcbcc31ed3e8e9cdff232badd0f18547971b, on @user_script:18.
```
when it tries to `SET ... GET` the key, since you can't `GET` a set.

This PR changes the `cache_thing_insideout.lua` script to [`redis.pcall()`](https://redis.io/docs/manual/programmability/lua-api/#redis.pcall) the `SET ... GET` and handle the error.

### DIFF
--- a/cacheops/lua/cache_thing_insideout.lua
+++ b/cacheops/lua/cache_thing_insideout.lua
@@ -15,7 +15,8 @@ end
 local stamps = {}
 local rnd = tostring(math.random())  -- A new value for empty stamps
 for _, conj_key in ipairs(conj_keys) do
-    local stamp = redis.call('set', conj_key, rnd, 'nx', 'get') or rnd
+    local stamp = redis.pcall('set', conj_key, rnd, 'nx', 'get')
+    if not stamp or stamp['err'] ~= nil then stamp = rnd end
     table.insert(stamps, stamp)
     -- NOTE: an invalidator should live longer than any key it references.
     --       So we update its ttl on every key if needed.


nicwolff:
Closed since I see issues – this'll need to delete the existing conj set or an insideout-style one will never be created.

Something needs to be done to make it possible to activate `CACHEOPS_INSIDEOUT` with existing conj sets in a production environment though...

Suor:
Why would you do that? You will either loose cache or loose invalidation information and get undead keys until they timeout. 

So error here is intentional, it shows that you can switch this setting without emptying your cache. Other issue is that data is encoded a little bit differently, i.e. a stamp is added.

Suor:
The issue of migrating cache is a valid one. This might be used on:
1. `CACHEOPS_INSIDEOUT` switch on or off.
2. `CACHEOPS_PREFIX` or `CACHEOPS_SERIALIZER` change.
3. Some backwards incompatible change in how `cacheops` codes things in Redis.

Solutions might be either a migration script or some way to read from one redis database in an old way and write to an another one the new way. 

You are welcome to create an issue for this.

PR #470, opened 2023-12-04
migrate from external mock package to stdlib unittest.mock

eli-schwartz:
It was moved to the stdlib in python 3.3, and the PyPI version is a backport for python versions less than 3.3.

Note: it was also never a dependendency so there was no guarantee it was installed to be used.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.9"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.10"]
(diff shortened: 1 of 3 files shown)


Suor:
And yet it somehow broken PyPy 3.9

eli-schwartz:
> And yet it somehow broken PyPy 3.9

Considering that PyPy 3.9 has segfaulted in the tests, I would very much like to know how my patch resulted in PyPy segfaulting.

I do not acknowledge that I broke anything. :) I don't know what happened, but I'm positive that it's not related.

Suor:
It might be their bug but it's my problem :) If I merge it I will need to fix it.

Suor:
Saying that, if changing from PyPy3.9 to 3.10 fixes it then it is fine with me.

eli-schwartz:
That appears to have magically fixed it.

Suor:
Thanks, merged

PR #477, opened 2024-06-10
[fix] conj keys always have -1 as expiration in redis 7

shubhamagrawal1906:
[fix] conj keys always have -1 as expiration in Redis 7
The issue in `redis.call('expire', conj_key, timeout, 'gt'` when setting up the value of ttl if its value is -1 then this function call won't work.


Suor:
Simply reverting to Redis 4 code should be both simpler and more efficient.

Suor:
Fixed in 6e37725f6f0a610006f9d20e55021445974ef65e

PR #488, opened 2024-11-25
Update md5 to set usedforsecurity for FIPS compliance

loganbibby:
For FIPS compliance, MD5 is not an allowed hashing algorithm when used for security. [In 3.9, `usedforsecurity` was introduced to flag non-security use of the algo.](https://docs.python.org/3/library/hashlib.html#hash-algorithms) 

This PR adds the `usedforsecurity` attribute to the instantiation with backwards compatibility for earlier Python versions.


cdunford:
This would be great to get merged

Issue #489, opened 2024-12-18
TypeError: Unhashable Type 'list' when using .exclude(some_array_field=[]) filter in Version 7.1


lozhkinandrei:
After updating to version `7.1`, we encountered a `TypeError: unhashable type: 'list'` when using the `.exclude(some_array_field=[])` filter. 

I will create a PR (https://github.com/Suor/django-cacheops/pull/490) with a test case that replicates this issue shortly to assist with debugging and resolution.

Thank you in advance for looking into this issue!

PR #490, opened 2024-12-18
test: add issue test #489

lozhkinandrei:
Add a test case for https://github.com/Suor/django-cacheops/issues/489


lozhkinandrei:
Note: have to set `CACHEOPS_DB` env var to `postgresql` in order for test to work

Suor:
Fixed it, thanks

PR #491, opened 2024-12-23
tweaks

raghavFam:


Suor:
These were dropped on purpose, not by random. You are welcome to use cacheops with older Pythons and Djangos but they won't be supported in future versions at least not explicitly.

PR #508, opened 2026-03-04
fix Subquery annotation detection in dnfs() for Django 6.0+

audriusm:
## Problem

Django 6.0 changed how `Subquery` expressions are stored in `qs.query.annotations`. In Django 5.x, annotation values remain as `Subquery` instances. In Django 6.0+, they are resolved into raw `django.db.models.sql.query.Query` objects during `.annotate()`.

The annotation dependency detection in `tree.dnfs()` only checks for `Subquery` instances:

```python
subqueries = (query_dnf(getattr(q, 'query', None))
              for q in qs.query.annotations.values() if isinstance(q, Subquery))
```
On Django 6, `isinstance(q, Subquery)` returns `False` for the resolved `Query` objects, so cross-table Subquery dependencies are completely invisible to cache invalidation.

This means any queryset with `Subquery` annotations on a cacheops-enabled model won't invalidate its cache when the subqueried table is modified - the cached result stays stale until timeout.

## How I found this

Came up while upgrading a project from Django 5.2 to 6.0. An endpoint annotating a cached model with a `Subquery` on a different table was returning stale data after writes to the subqueried table. Inspecting `dnfs()` output confirmed it only reported the primary table - the subqueried table was missing entirely.

## Fix

Check for both `Subquery` instances (Django <6) and resolved `Query` objects (Django 6+) when scanning annotations for table dependencies.

The existing `test_365 `covers this scenario and would fail on Django 6 without this fix.

Added `test_366` test to verify this fix.

### DIFF
--- a/cacheops/tree.py
+++ b/cacheops/tree.py
@@ -165,8 +165,7 @@ def dnfs(qs):
                 sub_dnfs.append(query_dnf(getattr(q, 'query', None)))
             elif isinstance(q, Query):
                 sub_dnfs.append(query_dnf(q))
-        if sub_dnfs:
-            dnfs_.update(join_with(lcat, sub_dnfs))
+        dnfs_.update(join_with(lcat, sub_dnfs) or {})
 
     return dnfs_
 


cacheops/tree.py
```diff
@@ -157,11 +157,18 @@ def table_for(alias):
...
-                      for q in qs.query.annotations.values() if isinstance(q, Subquery))
-        dnfs_.update(join_with(lcat, subqueries))
+        sub_dnfs = []
+        for q in qs.query.annotations.values():
+            if isinstance(q, Subquery):
+                sub_dnfs.append(query_dnf(getattr(q, 'query', None)))
+            elif isinstance(q, Query) and hasattr(q, 'model') and q.model is not None:
+                sub_dnfs.append(query_dnf(q))
+        if sub_dnfs:
+            dnfs_.update(join_with(lcat, sub_dnfs))
```
Suor: Please no circumstational changes like this.

cacheops/tree.py
```diff
@@ -157,11 +157,18 @@ def table_for(alias):
...
+    # so we check for both types.
     if qs.query.annotations:
-        subqueries = (query_dnf(getattr(q, 'query', None))
-                      for q in qs.query.annotations.values() if isinstance(q, Subquery))
-        dnfs_.update(join_with(lcat, subqueries))
+        sub_dnfs = []
+        for q in qs.query.annotations.values():
+            if isinstance(q, Subquery):
+                sub_dnfs.append(query_dnf(getattr(q, 'query', None)))
+            elif isinstance(q, Query) and hasattr(q, 'model') and q.model is not None:
```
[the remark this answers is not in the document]
audriusm: You're right, we don't  `Query.__init__` requires `model`, so it's always set. Removed the check in the latest push.

tests/tests.py
```diff
@@ -643,6 +643,30 @@ def test_365(self):
...
+        ).order_by('-pk').values('pk')
+        qs = Category.objects.cache().annotate(
+            newest_post=Subquery(newest_post[:1])
+        )
+        result = dnfs(qs)
+
+        self.assertEqual(result, {
+            Category._meta.db_table: [{}],
+            Post._meta.db_table: [{}],
+        })
```
Suor: Please change this to follow style aroud, i.e. less newlines

cacheops/tree.py
Suor: I mean don't need `if sub_dnfs:` you added here

audriusm: Removed the `if ..` but `join_with` returns `None` on empty input, so `or {}` is needed to keep this in single line

Suor:
Merged, thanks

PR #510, opened 2026-04-15
Fix #337 Bug when passing a query to execute that is a psycopg2 SQL object

### LINKED ISSUE #377: Bug when passing a query to execute that is a psycopg2 SQL object
When you pass a `SQL` object from `psycopg2.sql`, we run into an unexpected error in the [`is_sql_dirty` function](https://github.com/Suor/django-cacheops/blob/1524597c7645ebec6a5880d03c674a78fc210d19/cacheops/transaction.py#L108). When we hit the line that tries to do `sql.lower()`, we may have passed in a `SQL` object, resulting in ```packages/cacheops/transaction.py", line 113, in is_sql_dirty
    sql = sql.lower()
AttributeError: 'Composed' object has no attribute 'lower'```

 Perhaps it would be reasonable to add a block such as 

```python 
if isinstance(sql.SQL):
    sql = sql.as_string(self.connection)
```

Would love other thoughts on this.

gone:
Adds a fix to check the type of the incoming query and convert to string if it's not already one.
Fixes #377

### DIFF
--- a/cacheops/transaction.py
+++ b/cacheops/transaction.py
@@ -114,6 +114,9 @@ def is_sql_dirty(sql):
     # but some people will pass it anyway
     if isinstance(sql, bytes):
         sql = sql.decode()
+    # Handle psycopg2/psycopg3 sql.Composed/sql.SQL objects (see #377)
+    elif not isinstance(sql, str):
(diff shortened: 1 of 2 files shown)


Suor:
Thanks, merged

PR #513, opened 2026-09-10
Django 5.2 compatibility: prefetch related rely on TupleIn lookup

ddesvillechabrol:
Hello,
I experienced some issues with django 5.2 with prefetch related cache. Some prefetch related items in cache were not retrieved because the dnf part of conj keys was not set. Since django 5.2, the `get_prefetch_queryset` relies on Tuple Lookup since this commit https://github.com/django/django/commit/626d77e52a3f247358514bcf51c761283968099c .

The tuple lookup is used by composite primary keys newly introduced in django 5.2, sorry I don't have time to add every tuple lookup. So, I only added the TupleIn lookup that is used by prefetch related and I added a test with a composite primary key to check that TupleIn is taken into account by dnfs.

After the review, the composite key's conj key is set like `conj:tests_orderlineitem:product_id=1&order_id=A75H`
Prefetched items with unique keys keep the same conj ie: `conj:test_order:order_id=A75H`

Thanks !
Dimitri

### DIFF
--- a/cacheops/tree.py
+++ b/cacheops/tree.py
@@ -20,6 +20,12 @@ except ImportError:
     class SubqueryConstraint(object):
         pass
 
+try:
+    from django.db.models.fields.tuple_lookups import TupleIn
+except ImportError:
+    class TupleIn(object):
+        pass
+
 
 def dnfs(qs):
     """
@@ -46,6 +52,29 @@ def dnfs(qs):
         Any conditions other then eq are dropped.
         """
         if isinstance(where, Lookup):
+            if hasattr(where.lhs, 'targets'):
+                attnames = ','.join(t.attname for t in where.lhs.targets)
+                # since django5.2, prefetch_related rely on TupleIn and ColPair
+                if (
+                    isinstance(where, TupleIn)
+                    and len(where.rhs) < settings.CACHEOPS_LONG_DISJUNCTION
+                ):
+                    return {
+                        frozenset(
+                            {
+                                (
+                                    where.lhs.alias,
+                                    attnames,
+                                    ','.join(str(v) for v in values),
+                                    True,
+                                )
+                            }
+                        )
+                        for values in where.rhs
+                    }
+                else:
+                    # TODO: add other tuple lookup
+                    return SOME_TREE
             # If where.lhs don't refer to a field then don't bother
             if not hasattr(where.lhs, 'target'):
                 return SOME_TREE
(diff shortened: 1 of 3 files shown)


cacheops/tree.py
```diff
@@ -46,6 +52,29 @@ def _dnf(where):
...
+                if (
+                    isinstance(where, TupleIn)
+                    and len(where.rhs) < settings.CACHEOPS_LONG_DISJUNCTION
+                ):
+                    return {
+                        frozenset(
+                            {
+                                (
+                                    where.lhs.alias,
+                                    attnames,
```
Suor: I very much suspect this don't be invalidated properly. Need to simulate conj for these, i.e a == 1 & b == 2

cacheops/tree.py
```diff
@@ -1,14 +1,14 @@
...
+from django.db.models.lookups import Exact, In, IsNull, Lookup
 from django.db.models.query import QuerySet
 from django.db.models.sql import OR
 from django.db.models.sql.datastructures import Join
-from django.db.models.sql.query import Query, ExtraWhere
+from django.db.models.sql.query import ExtraWhere, Query
 from django.db.models.sql.where import NothingNode
-from django.db.models.lookups import Lookup, Exact, In, IsNull
-from django.db.models.expressions import BaseExpression, Exists
+from funcy import cat, group_by, join_with, lcat, lmap
```
Suor: Please no non-semantic changes

tests/models.py
```diff
@@ -308,6 +308,20 @@ def _curried(*moreargs, **morekwargs):
...
+class Product(models.Model):
+    name = models.CharField(max_length=100)
+
+
+class Order(models.Model):
+    reference = models.CharField(max_length=20, primary_key=True)
+
+
+class OrderLineItem(models.Model):
+    pk = models.CompositePrimaryKey("product_id", "order_id")
```
Suor: Should use if to make it work (be skipped) in older versions

Suor:
Hello, thanks for the PR. Can you please attend to the comments above. Also I see that tests are failing

ddesvillechabrol:
Yes I use feature only available for django 5.2+ so all test with django version before 5.2 failed and sorry for the auto formating 🙏 
I will check for invalidation with TupleIn. Thanks for your reactivity !

cacheops/tree.py
ddesvillechabrol: Yes you were right, the invalidation doesn't work. I tried something that work but the information is duplicated. Maybe I should take into account that the children is a tuple lookup to be able to transform it as an AND condition to have the conj like you said `a=1&b=2`. I will check that next monday I think

ddesvillechabrol: Hello ! I found some time today. The TupleIn conj is like the AND conj and it's well invalidated if the instance is updated :smile:
Tell me if you see any other issues :pray:

